PR #7, opened 2018-09-26
Use array.tobytes() on python3

microdog:
This can silence most of the deprecation warning.

### DIFF
--- a/thriftpy/protocol/compact.py
+++ b/thriftpy/protocol/compact.py
@@ -58,12 +58,12 @@ def write_varint(trans, n):
         else:
             out.append((n & 0xff) | 0x80)
             n = n >> 7
-    data = array.array('B', out).tostring()
+    data = array.array('B', out)
 
     if PY3:
-        trans.write(data)
+        trans.write(data.tobytes())
     else:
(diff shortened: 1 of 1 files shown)


thriftpy/protocol/compact.py
```diff
@@ -58,12 +58,12 @@ def write_varint(trans, n):
...
             n = n >> 7
-    data = array.array('B', out).tostring()
+    data = array.array('B', out)
 
     if PY3:
-        trans.write(data)
+        trans.write(data.tobytes())
     else:
-        trans.write(bytes(data))
+        trans.write(bytes(data.tostring()))
```
ethe: I think there are not any reasons to `bytes(data.tostring())`, maybe just `data.tostring()`, if param of function `bytes` is a string, it would return directly without process.

microdog: I am not sure what is the original intent to use `bytes` here, `bytes` should be an alias to `str` in python2. I will remove it.

ethe: It mights to adapt Python 3.x before.

PR #10, opened 2018-09-26
Fix travis build failed

ethe:

PR #11, opened 2018-09-26
Fix tests and travis build

microdog:
https://travis-ci.org/microdog/thriftpy2/builds/433592780

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,17 @@ python:
   - 3.4
   - 3.5
   - 3.6
-  - 3.7
   - pypy
 
 matrix:
   include:
     - python: 3.6
       script: tox -e flake8
-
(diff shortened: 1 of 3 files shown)


ethe:
No need to merge this any more.

microdog:
According to https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-401125783, the `3.7-dev` is using a fairly dated beta release that doesn't reflect the RC/stable 3.7, so it should not be used.

Also, there is a [behaviour change](https://bugs.python.org/issue32706#msg312895) in the ssl module which should be handled. I just updated the test to adopt this change, and I am not sure if the server or client code should be updated to adopt this.

ethe:
Please resolve conflicts, thanks! I don't think we need to revert #10

microdog:
Green. 🍻

PR #21, opened 2018-12-08
Parse IDL without sequence dependence

ethe:


farragar:
Thanks for fixing this! I'm no longer at the same company I was at when I raised https://github.com/Thriftpy/thriftpy/issues/280, but I've passed it along to a former colleague to check out :)

PR #37, opened 2019-01-28
Close socket when connect err

Junnplus:


ethe:
It works for me.

ethe:
@wooparadog @microdog

PR #38, opened 2019-02-12
Add i8 as alias for 'byte'

zhaakhi:
[THRIFT-3393](https://jira.apache.org/jira/browse/THRIFT-3393) added i8 as an alias for byte, for consistency with the other integer types.


ethe:
LGTM

PR #54, opened 2019-05-17
Include LICENSE.txt in sdist and wheel

xhochy:
It's always nice to ship the code with the license.


ethe:
Thank you, you are a nice guy.

Issue #48, opened 2019-03-11
Add URL parsing to make_client/client_context


dmulter:
Currently you must pass in `host`, `scheme`, and `path` when using the `HTTPClient`. It would be a nice feature to take a `url` parameter, and have the client method use `urllib.parse` to extract out these values for the caller.

ethe:
There is no standard DSN of Apache Thrift, but I think we can simply define it like `thrift://host:port`

Swixx:
@dmulter is the idea to have url as a kwarg?
@ethe does that mean we would have to change the signature to make `host` and `port` optional, but enforce one or the other?

ethe:
@Swixx I will not break down anything. URL will be an optional argument, you can still use host and port in the future.

Swixx:
@ethe I do not want to break anything either. I want to make a PR, but want to be sure that we get the best interface to the function possible. Currently, the signature is 
`make_client(service, host, port, path='', scheme='http',
                proto_factory=TBinaryProtocolFactory(),
                trans_factory=TBufferedTransportFactory(),
                ssl_context_factory=None,
                timeout=DEFAULT_HTTP_CLIENT_TIMEOUT_MS):`

If we put an optional kwarg for `url`, that essentially makes host and port unnecessary, but they are currently required positional arguments, so a call such as
`make_client(pingpong.PingService, url='https://example.com:443', ssl_context_factory=ssl_context_factory)`
would be invalid, even though we could extract out the necessary information.

What would your idea be for how to call the function if we made `url` a kwarg?

We could overload the function, but I don't know if that is necessary.

ethe:
@Swixx I think
```python
def make_client(service, host='', port='', path='', scheme='http', proto_factory=TBinaryProtocolFactory(), trans_factory=TBufferedTransportFactory(), ssl_context_factory=None, timeout=DEFAULT_HTTP_CLIENT_TIMEOUT_MS, url=''):
    pass
```
is a good solution. Make host and port optional and put the url end of the args will not break anything, user can still use `make_client(pingpong.PingService, '127.0.0.1', '5000')`, and also `make_client(pingpong.PingService, url='http://127.0.0.1:5000')`

Swixx:
@ethe there is currently a `make_client` in `http.py`, `rpc.py`, `tornado.py`, and `contrib/aio/rpc.py`. Should all of these implement the new functionality?

[...]

PR #57, opened 2019-05-24
Added optional  argument to make_client() definition to allow for url…

Swixx:
… parsing.

#48.

### DIFF
--- a/tests/test_framed_transport.py
+++ b/tests/test_framed_transport.py
@@ -83,7 +83,7 @@ class FramedTransportTestCase(TestCase):
 
     def mk_client(self):
         return make_client(addressbook.AddressBookService,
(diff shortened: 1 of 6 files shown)


Swixx:
@ethe I'm willing to go back and try to add test cases, but make_client is only used in two places (`teset_framed_transport.py` and `test_tornado.py`), and they're used by a wrapper function in both places (`mk_client()`). What's the best way to add a test case for the newly added `url` portion, given that it isn't truly tested now, but just used in the `setUp`?

ethe:
I checked your pull request, the current way of adding test case is fine, but please fix the failed reason of CI. Even it is ran in setUp, it can still promise the make_client function works as expected.

ethe:
Closed by #80

PR #60, opened 2019-05-30
Add default timeout for make_client

aisk:
I think it's necessary for having a default timeout for `make_client`, and the `make_server` / `make_aio_server` / `make_aio_client` have the default timeout already.


ethe:
Thank you, I agree the behavior of make_aio_client and make_client should have the same default timeout settings.

PR #62, opened 2019-05-31
Fixes #61 parse error on empty set field value

### LINKED ISSUE #61: Parse error on empty set field value
The following thift file,

(**NB** that the file can be parsed correctly using the Thrift executable from https://thrift.apache.org/download,  tested with the latest version at the time of writing 0.12.0)

```thrift
struct Status {
	1: string state,
	2: optional set<string> info={},  # <-- problem line
}
	
service EchoService {
    string echo(1: string param),
    Status get_status(),
}
```

causes a parse error,
```python
Traceback (most recent call last):
  File "C:\Users\daniel.farrell\AppData\Local\Continuum\anaconda3\envs\porosity_env\lib\site-packages\thriftpy2\parser\parser.py", line 341, in p_simple_field
    val = _cast(p[3])(p[6])
  File "C:\Users\daniel.farrell\AppData\Local\Continuum\anaconda3\envs\porosity_env\lib\site-packages\thriftpy2\parser\parser.py", line 753, in __cast_set
    assert isinstance(v, (list, set))
AssertionError
```

The parse error is because Python does not have a literal for an empty set. The string `{}` in the thrift file is interpreted as a dict which generates the assertion error.

https://github.com/Thriftpy/thriftpy2/blob/cbc250d0d4473c20e9de08c3eea6cc44cf01c393/thriftpy2/parser/parser.py#L751

A potential fix is to check for empty dict values and force the return value to be a set:

```python
def _cast_set(t):
    assert t[0] == TType.SET

    def __cast_set(v):
        if len(v) == 0 and isinstance(v, dict):
            v = set()
        assert isinstance(v, (list, set))
        map(_cast(t[1]), v)
        if not isinstance(v, set):
            return set(v)
        return v
    return __cast_set
```

I'm working on a pull request and adding tests etc.

danieljfarrell:
* Added a check in the parser to convert lines with empty set constants.
 * To test this followed the example of the email struct and added a metadata struct
  to the email which holdes an empty set of strings. Seemed like the most logical way
  to include parser bug in the existing tests.

### DIFF
--- a/tests/parser-cases/structs.thrift
+++ b/tests/parser-cases/structs.thrift
@@ -3,14 +3,21 @@ struct Person {
     2: string address
 }
 
+struct MetaData {
+    1: set<string> tags = {}
+}
+
 struct Email {
     1: string subject = 'Subject',
     2: string content,
     3: Person sender,
     4: required Person recver,
+    5: MetaData metadata
 }
+
 
 const Email email = {'subject': 'Hello', 'content': 'Long time no see',
     'sender': {'name': 'jack', 'address': 'jack@gmail.com'},
-    'recver': {'name': 'chao', 'address': 'chao@gmail.com'}
+    'recver': {'name': 'chao', 'address': 'chao@gmail.com'},
+    'metadata': {},
 }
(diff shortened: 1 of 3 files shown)


ethe:
I merged this, thank you.

PR #63, opened 2019-06-04
enable include_package_data in setup.py

aisk:
The original problem is the `.pyx` and `.pxd` files was not packaged in current versions, so it's not possible for extending the cython based classes (cython need source files to compile) in another package besides thriftpy2.

I saw there was a `MANIFEST.in`,but the `include_package_data` is not set in `setup.py`, according to 
[setuptoos 's documents](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files), it should be set to make `MANIFEST` to take efforts. And seems the `package_data` field is duplicated if we enable `include_package_data`.


ethe:
If you want to package thriftpy2 manually, using `make build` instead, mentioned that you need to install `Cython` correctly. I think `make build`has already processed the `.pyx` and `.pxd` file.

aisk:
No, I want using thriftpy2 as a pip dependence, and extend some classes which was written in cython.

aisk:
And I think it's a bug that `MANIFEST.in` was there and take no efforts.

ethe:
Could you please tell me how can I reproduce your issue?

aisk:
Running `pip install thriftpy2` to install latest thriftpy2, and take a look and site-packages, the `pyx` files was gone, only the `.so` exists. If we want subclass a cython class, the `pyx` must exists.

Install my branch using `pip install --user git+https://github.com/aisk/thriftpy2.git#branch=fix-package-data`, the `pyx` exists as expected.

ethe:
That's true, thank you.

aisk:
Hi @ethe , another question, what's the release schedule of thriftpy2? My package is reply on this and #60.

ethe:
@aisk I will release it later this week, unfortunately there is not a release schedule of thriftpy2 yet.

aisk:
Cool, thanks for help!

PR #67, opened 2019-06-16
Support kwargs style parameters passing in TSimpleServer

aisk:
`kwargs` is supported in `TThreadedServer`, so I think `TSimpleServer` should support it too.


ethe:
Yeah it should be, thank you!

PR #80, opened 2019-08-31
add url support as optional argument to make_client

iamsudip:


ethe:
I think it is better to change the protocol type in the URL, `https://` and `tcp://` does not make sense. In my opinion, using `thrift://` is better because of the usual practice.

iamsudip:
> I think it is better to change the protocol type in the URL, `https://` and `tcp://` do not make sense. In my opinion, using `thrift://` is better because of the usual practice.

Agree. Let me change that.

iamsudip:
@ethe Have a look now! Made the changes and added more tests to increase coverage ;)

ethe:
Merged, thank you.

PR #81, opened 2019-08-31
enforce required arguments, fixes #72

### LINKED ISSUE #72: Thriftpy should enforce required arguments
It seems that thriftpy is not enforcing required parameters to service calls. Instead of the client raising an exception, the request is serialized with `None` in place of the missing arguments. I could see this being reasonable for optional arguments, but required arguments should be required.

For the example service:

```thrift
service HelloService {
    string greet(1: required string greeting);
}
```

Calling `client.greet()` will send `None` as the value for `greeting`. I would expect this to throw an exception.

iamsudip:


iamsudip:
@ethe Can you have a look? It doesn't validate argument type though.

ethe:
Thank you! BTW, could you mind to add some test cases to ensure the behavior of your enhancement?

thriftpy2/thrift.py
```diff
@@ -13,12 +13,22 @@
...
+if PY3:
+    from itertools import zip_longest
+else:
+    from itertools import izip_longest as zip_longest
 
 
-def args2kwargs(thrift_spec, *args):
-    arg_names = [item[1][1] for item in sorted(thrift_spec.items())]
-    return dict(zip(arg_names, args))
+def args2kwargs(thrift_spec, *args, **kwargs):
```
ethe: `args_to_kwargs` is better than `args2kwargs` I think.

iamsudip: Done

ethe:
I think it is fine, merged it. The client does not check the type of arguments.

PR #83, opened 2019-09-09
Enable socket_timeout on unix_socket

ryerh:
The `timeout` param for `rpc.make_client` only works when user provides `host:port`, but ignored when user provides `unix_socket`.

This PR fixed it.

Could you bump a new version as soon as this PR is accepted? pls.


ethe:
LGTM, thank you!

PR #85, opened 2019-09-20
protocol: follow strict datatype in TJsonProtocol, fixes #71

### LINKED ISSUE #71: Follow Strict Data Type in TJsonProtocol
Hey, I hava met a problem about using TJsonProtocol to serialize thrift obejct.

There is a double field defined in my thrift struct but assigned with a decimal object, which can not be serialized with TJsonProtocol.

I found the json_value function just return the value if it is a double field.

    def json_value(ttype, val, spec=None):
        if ttype in INTEGER or ttype in FLOAT or ttype == TType.STRING:
            return val
    
        if ttype == TType.BOOL:
            return True if val else False
    
        if ttype == TType.STRUCT:
            return struct_to_json(val)
    
        if ttype in (TType.SET, TType.LIST):
            return list_to_json(val, spec)
    
        if ttype == TType.MAP:
            return map_to_json(val, spec)

I know the easiest way is to change the decimal object to float object, but it is possible to follow the strict data type in json value? such as:

    def json_value(ttype, val, spec=None):
        if ttype in INTEGER:
    	   return int(val)

        if ttype in FLOAT:
    	   return float(val)

        if ttype == TType.STRING:
    	   return str(val)
    
        if ttype == TType.BOOL:
            return True if val else False
    
        if ttype == TType.STRUCT:
            return struct_to_json(val)
    
        if ttype in (TType.SET, TType.LIST):
            return list_to_json(val, spec)
    
        if ttype == TType.MAP:
            return map_to_json(val, spec)

Thanks a lot!

iamsudip:


thriftpy2/protocol/json.py
```diff
@@ -5,51 +5,50 @@
...
-        return int(val)
-
-    if ttype in FLOAT:
-        return float(val)
-
-    if ttype in (TType.STRING, TType.BOOL):
-        return val
-
-    if ttype == TType.STRUCT:
-        return struct_to_obj(val, spec())
```
iamsudip: @ethe I'm confused here.

In this version `spec` is getting called. In my version if I call, tests fail; if I don't, tests pass.

thriftpy2/protocol/json.py
```diff
@@ -5,51 +5,50 @@
...
-        return map_to_obj(val, spec)
+    TTYPE_TO_OBJFUNC_MAP = {
+        TType.BYTE: (int, (val, )),
+        TType.I16: (int, (val, )),
+        TType.I32: (int, (val, )),
+        TType.I64: (int, (val, )),
+        TType.DOUBLE: (float, (val, )),
+        TType.STRING: (u, (val, )),
+        TType.BOOL: (bool, (val, )),
+        TType.STRUCT: (struct_to_obj, (val, spec)),
```
iamsudip: @ethe I didn't call `spec`.

thriftpy2/protocol/json.py
ethe: Cause spec is called only a single time in the old version, but it would be called multiple times if using dict rather than if branch.

thriftpy2/protocol/json.py
ethe: Please do not modify it.

thriftpy2/protocol/json.py
```diff
@@ -5,53 +5,57 @@
...
-
-    if ttype == TType.STRUCT:
-        return struct_to_json(val)
-
-    if ttype in (TType.SET, TType.LIST):
-        return list_to_json(val, spec)
-
-    if ttype == TType.MAP:
-        return map_to_json(val, spec)
+    TTYPE_TO_JSONFUNC_MAP = {
```
iamsudip: @ethe Do you think with there is negative performance impact if we move to map from branching?

thriftpy2/protocol/json.py
```diff
@@ -5,53 +5,57 @@
...
-    if ttype in INTEGER:
-        return int(val)
-
-    if ttype in FLOAT:
-        return float(val)
-
-    if ttype in (TType.STRING, TType.BOOL):
-        return val
-
+    TTYPE_TO_OBJFUNC_MAP = {
```
iamsudip: Same here.

thriftpy2/protocol/json.py
ethe: I think there is not much impact here.

PR #91, opened 2019-10-08
Extend test cert validity to 2049

bmwiedemann:
Extend test cert validity to 2049
to make tests pass in 2022

Background:
As part of my work on reproducible builds for openSUSE, I check
that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect
some software will be used in some places.
This showed up failing tests in our package build.

See https://reproducible-builds.org/ for why this matters.

Note: I tested that i586 builds still pass tests on 2037-12-01
so are not affected by the year 2038 bug yet.


ethe:
LGTM, thank you!

PR #101, opened 2019-11-27
Fix async read size bugs

aiudirog:
- Make sure TAsyncSocket reads all of the data
- Ask for the min size, not the max in TAsyncBufferedTransport.read()

### DIFF
--- a/thriftpy2/contrib/aio/socket.py
+++ b/thriftpy2/contrib/aio/socket.py
@@ -183,9 +183,13 @@ class TAsyncSocket(object):
                 # in lib/cpp/src/transport/TSocket.cpp.
                 self.close()
                 # Trigger the check to raise the END_OF_FILE exception below.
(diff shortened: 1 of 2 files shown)


ethe:
Do not forget to handle `IncompleteReadError`, re-raise it as `TTransportException.END_OF_FILE` is a good idea.

aiudirog:
Good catch, thanks!

ethe:
Please see #102, this part of code was not written by me so I spent a lot of time to understand it, I think it may conform the origin design. The key is, using `min` would lose the cache from buffer.

ethe:
Maybe you can test your cases with my branch, thank you.

aiudirog:
I'll look into it in the morning, thanks for the quick responses!

aiudirog:
Closed in favor of #102

PR #102, opened 2019-11-27
Fix unexpected data length in aio buffer transport

ethe:


aiudirog:
It passes my tests with one change: `rest_len = len(ret) - sz` should be `rest_len = sz - len(ret)` since sz will be bigger.

However, if you use max for the read length, TAsyncSocket can't use readexactly otherwise we're back to the timeout errors. This means TAsyncSocket has to be wrapped in the buffered transport and couldn't be used as transport on its own because it's read method doesn't guarantee the number of bytes. If that's okay, then this PR should be it.

ethe:
Yes, we should use `read` rather than `readexactly` here.

aiudirog:
Yeah, looking at TSocket it does the same thing. Sounds good! I'll close the other PR.

PR #103, opened 2019-12-02
Add TAsyncCompactProtocol and TAsyncFramedTransport

aiudirog:
For consistency with the standard implementation, I've added the async versions of the compact protocol and framed transport. 

I had to refactor the aio test cases because they were getting unruly. I chose to have a [base test case](https://github.com/Thriftpy/thriftpy2/blob/195811125b2b34c97ec1b08869ff965ca69cca2d/tests/test_aio.py#L108) which has all of the generic tests and then [a subclass for each combination](https://github.com/Thriftpy/thriftpy2/blob/195811125b2b34c97ec1b08869ff965ca69cca2d/tests/test_aio.py#L258) of protocol, transport, and SSL ([the SSL is added as a mixin](https://github.com/Thriftpy/thriftpy2/blob/195811125b2b34c97ec1b08869ff965ca69cca2d/tests/test_aio.py#L216)). If you don't like this design I can find a different way.

-------------------------

I also found a very small bug in the compact protocol where booleans with field id 0 weren't being written correctly. As you can see in the code below, the Java implementation checks to see if the bool fid is not null and if so calls their write field header (`writeFieldBeginInternal()`) where it then checks if it can compact with previous field.  The Python code checks if the bool fid is falsey as well as some of the internal conditions for compacting the fields which prevents it from calling `_write_field_header()` when it should. 

[Python Source](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/protocol/compact.py#L402)
```python
    def write_bool(self, bool):
        # Use a ternary once instead of separate if statements
        # Changed to make this example easier to read
        ctype = CompactType.TRUE if bool else CompactType.FALSE
        if self._bool_fid and self._bool_fid > self._last_fid \
                and self._bool_fid - self._last_fid <= 15:
            self._write_field_header(ctype, self._bool_fid)
        else:
            self.write_byte(ctype)

    def _write_field_header(self, type, fid):
        delta = fid - self._last_fid
        if 0 < delta <= 15:
            self.write_ubyte(delta << 4 | type)
        else:
            self.write_byte(type)
            self.write_i16(fid)
        self._last_fid = fid
```

[Java Source](https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/protocol/TCompactProtocol.java#L314)
```java
  public void writeBool(boolean b) throws TException {
    if (booleanField_ != null) {
      // we haven't written the field header yet
      writeFieldBeginInternal(booleanField_, b ? Types.BOOLEAN_TRUE : Types.BOOLEAN_FALSE);
      booleanField_ = null;
    } else {
      // we're not part of a field, so just write the value.
      writeByteDirect(b ? Types.BOOLEAN_TRUE : Types.BOOLEAN_FALSE);
    }
  }
  private void writeFieldBeginInternal(TField field, byte typeOverride) throws TException {
    // short lastField = lastField_.pop();

    // if there's a type override, use that.
    byte typeToWrite = typeOverride == -1 ? getCompactType(field.type) : typeOverride;

    // check if we can use delta encoding for the field id
    if (field.id > lastFieldId_ && field.id - lastFieldId_ <= 15) {
      // write them together
      writeByteDirect((field.id - lastFieldId_) << 4 | typeToWrite);
    } else {
      // write them separate
      writeByteDirect(typeToWrite);
      writeI16(field.id);
    }

    lastFieldId_ = field.id;
    // lastField_.push(field.id);
  }
```

If the original intention was to only call `_write_field_header()` if the compaction was going to happen, this should make it work (but I think the Java implementation is more efficient as it doesn't compute the delta twice):
```python
    def write_bool(self, bool):
        ctype = CompactType.TRUE if bool else CompactType.FALSE
        if self._bool_fid and self._bool_fid > self._last_fid \
                and self._bool_fid - self._last_fid <= 15:
            self._write_field_header(ctype, self._bool_fid)
        else:
            self.write_byte(ctype)
            if self._bool_fid is not None:
                self.write_i16(self._bool_fid)
```

If you leave off the bug fix commit, you should see that my test cases fail when the protocol needs to return a boolean.

### DIFF
--- a/thriftpy2/contrib/aio/protocol/compact.py
+++ b/thriftpy2/contrib/aio/protocol/compact.py
@@ -3,7 +3,7 @@
 from __future__ import absolute_import
 
 import asyncio
-from struct import pack, unpack
+from struct import unpack
 
 from thriftpy2.protocol.exc import TProtocolException
 from thriftpy2.thrift import TException, TType
--- a/thriftpy2/contrib/aio/protocol/compact.py
+++ b/thriftpy2/contrib/aio/protocol/compact.py
@@ -64,7 +64,7 @@ class TAsyncCompactProtocol(TCompactProtocol):  # Inherit all of the writing
         return name, type, seqid
 
     @asyncio.coroutine
-    def read_message_end(self):
+    def read_message_end(self):  # TAsyncClient expects coroutine
         assert len(self._structs) == 0
 
     @asyncio.coroutine
@@ -88,16 +88,13 @@ class TAsyncCompactProtocol(TCompactProtocol):  # Inherit all of the writing
 
         return None, self._get_ttype(type), fid
 
-    @asyncio.coroutine
     def read_field_end(self):
         pass
 
-    @asyncio.coroutine
     def read_struct_begin(self):
         self._structs.append(self._last_fid)
         self._last_fid = 0
 
-    @asyncio.coroutine
     def read_struct_end(self):
         self._last_fid = self._structs.pop()
 
@@ -120,7 +117,6 @@ class TAsyncCompactProtocol(TCompactProtocol):  # Inherit all of the writing
             size = yield from self._read_size()
         return type, size
 
-    @asyncio.coroutine
     def read_collection_end(self):
         pass
 
@@ -166,7 +162,7 @@ class TAsyncCompactProtocol(TCompactProtocol):  # Inherit all of the writing
 
     @asyncio.coroutine
     def read_struct(self, obj):
-        yield from self.read_struct_begin()
+        self.read_struct_begin()
         while True:
             fname, ftype, fid = yield from self.read_field_begin()
             if ftype == TType.STOP:
@@ -189,8 +185,8 @@ class TAsyncCompactProtocol(TCompactProtocol):  # Inherit all of the writing
                     setattr(obj, fname, val)
                 else:
                     yield from self.skip(ftype)
-            yield from self.read_field_end()
-        yield from self.read_struct_end()
+            self.read_field_end()
+        self.read_struct_end()
 
     @asyncio.coroutine
     def read_val(self, ttype, spec=None):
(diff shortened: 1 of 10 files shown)


ethe:
Thank you, this is awesome, I am reviewing it/

thriftpy2/contrib/aio/protocol/compact.py
```diff
@@ -0,0 +1,318 @@
...
+    def read_field_end(self):
+        pass
+
+    @asyncio.coroutine
+    def read_struct_begin(self):
+        self._structs.append(self._last_fid)
+        self._last_fid = 0
+
+    @asyncio.coroutine
+    def read_struct_end(self):
```
ethe: Seems like there are some methods do not need to add `asyncio.coroutine` decorator.

ethe:
I think the refactored test cases are pretty well, thank you.

thriftpy2/contrib/aio/protocol/compact.py
aiudirog: I decorated all the methods starting with read to make it easier to convert by assuming all reads were coroutines. I can change it if you like as there is a slight performance cost.

ethe: Yes, if the method does not depend on I/O, it is better to remove the coroutine decorator.

aiudirog: I removed all the ones I could. TAsyncClient assumes `read_message_end()` is a coroutine so I left that one with the decorator.

aiudirog:
While running the tests again, I noticed some warnings that showed the clients and servers weren't being cleaned up properly. I updated the test cases to ensure all clients are closed and I fixed a bug in `TAsyncServer.serve()` where `close()` wasn't being awaited.

thriftpy2/contrib/aio/protocol/compact.py
aiudirog: It may be worth it to go through all the different protocol classes and mark any helper methods as private so it's easier to determine which need to be async regardless of content based upon the public interface.

ethe: Yes, maybe we should promise an interface to describe which methods of a protocol should be coroutines.

aiudirog: I can take care of that in a dedicated PR for standardizing interfaces.

PR #106, opened 2019-12-05
contrib.aio.processor: Fix missing asyncio.coroutine decorator

achimnol:
This PR fixes an error `TypeError: object generator can't be used in 'await' expression` when I call thriftpy2 on latest Python 3.8 with native async functions.


ethe:
LGTM, thank you!

PR #107, opened 2019-12-06
Support nest incomplete type

### LINKED ISSUE #105: Invalid thrift_spec if type is used before define
Codes to reproduce this:

test.thrift

```thrift
struct Container {
    1: A field1;
    2: list<A> field2;
    3: list<map<string, A>> field3;
}

struct A {
    1: string value
}
```

test.py

```python
import thriftpy2
from pprint import pprint

test_thrift = thriftpy2.load('./test.thrift')

pprint(test_thrift.Container.thrift_spec)
```

output:

```
{1: (12, 'field1', <class 'test.A'>, False),
 2: (15, 'field2', (12, <class 'test.A'>), False),
 3: (15, 'field3', (13, (11, -3)), False)}
```

Notice the `field3`, it's spec is expected to be `(15, 'field3', (13, (11, <class 'test.A'>)), False)` like `field2`.

aisk:
Try to fix #105 .

### DIFF
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -111,6 +111,20 @@ def fill_incomplete_ttype(tmodule, definition):
                         )
                     ),
                     value[3])
+            # if the field's ttype is a nest compound type
+            # and it contains incomplete type
+            elif isinstance(value[2], tuple):
+                def walk(part):
+                    if isinstance(part, tuple):
+                        return tuple(walk(x) for x in part)
+                    if part in incomplete_type:
+                        return get_definition(tmodule, *incomplete_type[part])
+                    return part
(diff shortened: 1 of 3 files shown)


thriftpy2/parser/__init__.py
```diff
@@ -111,6 +111,18 @@ def fill_incomplete_ttype(tmodule, definition):
                         )
                     ),
                     value[3])
+            elif isinstance(value[2], tuple):
```
ethe: It is better to leave some comments here.

ethe:
`fill_incomplete_ttype ` has become too hard to understand, it is better to leave some comments there, maybe I will refactor it in the future.

aisk:
@ethe updated

ethe:
LGTM

PR #108, opened 2019-12-07
Add TAsyncProtocolBase and TAsyncTransportBase

aiudirog:
As promised, I'm implementing proper interfaces and cleaning up some parts of the code. I'll let you know when this is ready.:)

### DIFF
--- a/thriftpy2/transport/__init__.py
+++ b/thriftpy2/transport/__init__.py
@@ -4,92 +4,13 @@ from __future__ import absolute_import
 
 from thriftpy2._compat import CYTHON
 
-from ..thrift import TType, TException
-
-
-def readall(read_fn, sz):
-    buff = b''
-    have = 0
-    while have < sz:
-        chunk = read_fn(sz - have)
-        have += len(chunk)
-        buff += chunk
-
-        if len(chunk) == 0:
-            raise TTransportException(TTransportException.END_OF_FILE,
-                                      "End of file reading from transport")
-
-    return buff
-
-
-class TTransportBase(object):
-    """Base class for Thrift transport layer."""
-
-    def is_open(self):
-        """Check if this transport is open."""
-        raise NotImplementedError
-
-    def open(self):
-        """
-        Prepare this transport for usage and allocate any necessary resources
-        like sockets or sessions.
-        """
-        raise NotImplementedError
-
-    def close(self):
-        """Clean up and deallocate any resources allocated in open()."""
-        raise NotImplementedError
-
-    def _read(self, sz):
-        """
-        Internal read method which can read up to `sz` bytes but doesn't
-        need to return them all.
-        """
-        raise NotImplementedError
-
-    def read(self, sz):
-        """Get exactly `sz` bytes from the underlying connection."""
-        return readall(self._read, sz)
-
-    def write(self, buf):
-        """
-        Submit some data to tbe written to the connection. May be
-        buffered until flush is called.
-        """
-        raise NotImplementedError
-
-    def flush(self):
-        """Ensure that all internal buffers are emptied into the connection."""
-        raise NotImplementedError
-
-
-class TTransportException(TException):
-    """Custom Transport Exception class"""
-
-    thrift_spec = {
-        1: (TType.STRING, 'message'),
-        2: (TType.I32, 'type'),
-    }
-
-    UNKNOWN = 0
-    NOT_OPEN = 1
-    ALREADY_OPEN = 2
-    TIMED_OUT = 3
-    END_OF_FILE = 4
-
-    def __init__(self, type=UNKNOWN, message=None):
-        super(TTransportException, self).__init__()
-        self.type = type
-        self.message = message
-
-
-# Avoid recursive import
-from .socket import TSocket, TServerSocket  # noqa
-from .sslsocket import TSSLSocket, TSSLServerSocket  # noqa
-from ._ssl import create_thriftpy_context  # noqa
-from .buffered import TBufferedTransport, TBufferedTransportFactory  # noqa
-from .framed import TFramedTransport, TFramedTransportFactory  # noqa
-from .memory import TMemoryBuffer  # noqa
+from .base import TTransportBase, TTransportException
+from .socket import TSocket, TServerSocket
+from .sslsocket import TSSLSocket, TSSLServerSocket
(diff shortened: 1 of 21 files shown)


thriftpy2/transport/__init__.py
```diff
@@ -25,12 +25,43 @@ def readall(read_fn, sz):
...
     def _read(self, sz):
+        """
+        Internal read method which can read up to `sz` bytes but doesn't
+        need to return them all.
+        """
         raise NotImplementedError
 
     def read(self, sz):
+        """Get exactly `sz` bytes from the underlying connection."""
         return readall(self._read, sz)
```
aiudirog: The original `TTransportBase` seems to declare that `read()` will always return `sz` bytes. However not every transport on its own does this. `TFramedTransport` is a good example because it always assumes it will be wrapped by `TBufferedTransport` and would therefore time out if it tried to read everything it was asked to. Might it be best to have the protocol call `readall()` instead? Then we would remove `_read()` and make `read()` an abstract method.

If we decide not to make that change, I'll update this doc-string to say that if `read()` isn't going to return all the requested bytes, the transport must be wrapped by one that will.

aiudirog:
@ethe I'm all set for now. Let me know what you think and if there is anything else you want to add.

thriftpy2/transport/__init__.py
ethe: I think read must return `sz` bytes, that is the reason I modified the TAsyncBufferedTransport

ethe: If transport can not find a way to return enough length data, it should raise a `END_OF_FILE` exception.

ethe: I have no idea about `readall` method, could you please tell me which cases are this method used?

aiudirog: The socket should always be able to furnish the data requested by the protocol, otherwise EOF would get raised. However transports are designed to wrap each other to form layers and, internally, calls to `read()` between wrapped transports can't always return `sz` bytes. 

Like we discussed previously, `TSocket` has to be wrapped by `TBufferedTransport` in order to be used as a transport despite it sporting the same API. However, because of the nature of `TBufferedTransport` and it requesting as much data as possible from the sub-transport to fill its buffer, the read method of the sub-transport can't always return `sz` bytes. This is the case with `TFramedTransport` which, despite inheriting the [readall version of `read()` from `TTransportBase`](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/transport/__init__.py#L31), does not implement [`_read()`](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/transport/__init__.py#L28) (to the dismay of most linters:P) and instead [overrides `read()`](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/transport/framed/__init__.py#L29) so that it doesn't have to return all `sz` bytes. This allows it to be wrapped by `TBufferedTransport` without running into timeout errors when it can't furnish enough data to fill the buffer but it also makes the wrapping required to ensure all the data is returned. 

This leads to a weird dynamic where a transport has to make one of two choices:
1. Have its `read()` return exactly `sz` bytes and not support being wrapped by `TBufferedTransport`
2. Have its `read()` return whatever is in the underlying transport and require being wrapped by `TBufferedTransport`

Basically, the current implementation requires that the outermost transport has to return exactly `sz` bytes from its `read()` and that inner transport just returns what it can get its hands on and will be called again if it didn't get enough. The means that transports like `TFramedTransport` can't be put on the outer layer and anyone developing a custom transport needs to know the rules above, which aren't obvious from looking at the base class.

I propose we could make this easier one of two ways:
- Have the protocol be in charge of calling `readall()` instead of `TBufferedTransport`. 
     - This would require quite a few changes in places and the more I look at it, the more I don't like it.
- Have two public methods for reading: `read()` which returns exactly `sz` bytes and `read1()` which returns up to `sz` bytes.
    - This matches the standard Python Buffer API from [io.IOBase](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read)
        - I don't particularly like the name `read1`, but I chose it since it matches the standard API. We could always change it to something like `read_up_to`.
    - `read1()` would be used for intra-transport reads like `TBufferedTransport` to `TFramedTransport` and `read()` would be used by the protocol to guarentee all of the data gets read.
    - Implementation would be quite easy: the existing `_read()` method would become `read1()` and serve the same purpose, only publicly. `TTransportBase.read()` would stay exactly as is and be inherited by every transport so they would only have to define a `read1()`.

Either change would allow transports to wrap each other arbitrarily, making it easier to develop new, more complex transports in the future.

I'll mock up a commit of the second one and send it over. We can always drop it.

aiudirog: https://github.com/aiudirog/thriftpy2/commit/d2bcc63a441e24d25fdab9a4064387613157e73a

It's not 100% perfect (I don't like the sockets breaking the read1 vs read) but it passes the test cases and makes transport implementation more consistent.

ethe: I checked the exists logic, I got a mistake, Transport does not always return the data equal to the size exactly, This feature was caused by TSocket and Transport just inherit it. I am still trying to understand the rest part of your description.

ethe: I think I have got your point. `TFramedTransport` does not need to return data exactly equal to the required size, but `TBufferedTransport` does. However, I think it is better to keep this behavior as before, leave some comments to the `read` method to describe this behavior is fine I think.

ethe: I do not like the design of `read` and `read1` or `read_up_to`, but it seems like replace `read` to `readall` would break the compatibility

aiudirog: > leaving some comments to the `read` method to describe this behavior is fine I think.

Sounds good. As long as it's clearly documented it shouldn't be a problem.

aiudirog: I wanted to see how Java did it and they actually have a `read()` and `readall()`: https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/transport/TTransport.java#L60

I'll update the documentation for now though.

aiudirog:
I updated the docstring and included a link to our discussion. Since you approved the previous changes, I'll go ahead and merge it in. Thanks!

PR #114, opened 2020-01-13
Include tests in the sdist tarball on pypi

scarabeusiv:
If a vendor wants to validate the thriftpy2 against his python stack this allows him to use the pypi archive instead of relying on github tags.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ include README.rst CHANGES.rst LICENSE
 recursive-include thriftpy2/protocol/cybin *.pyx *.c *.h
 recursive-include thriftpy2/transport *.pyx *.pxd *.c
 include thriftpy2/contrib/tracking/tracking.thrift
+recursive-include tests/ *


ethe:
I checked other libraries such as flask and redis-py, they do not pack their test cases in their package, could you please tell me your primary demand?

scarabeusiv:
We use this to execute tests on distribution side, i.e. the package that Debian/Fedora/SUSE ships we take from pypi and also execute the tests to see if it works.

See example buildlog of pytest-testconfig package: [pytest-buildlog.txt](https://github.com/Thriftpy/thriftpy2/files/4054319/pytest-buildlog.txt)

aiudirog:
The tests package is already excluded `setup.py` under `find_packages` so the manifest change would only include them in the sdist. To me, that makes sense because anyone building from source is probably going to want to run the tests, but someone installing a wheel probably won't.

ethe:
I got it, thank you.

PR #115, opened 2020-01-18
Rename socket_timeout to timeout for compatibility

aiudirog:
The [standard `make_client()`](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/rpc.py#L33) uses the `timeout` argument to specify the socket timeout whereas the [asyncio `make_client()`](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/contrib/aio/rpc.py#L26) calls it `socket_timeout`. Since AIOHappyBase handles both, it leads to awkward dispatch issue where I have to rewrite the argument for one. 

This PR deprecates `socket_timeout` in favor of just `timeout` so that the async `make_client()` matches the standard one. I chose this and not the other way around so it would have the least impact on existing code, even though the `socket_timeout` name is more descriptive and probably better.

### DIFF
--- a/tests/test_aio.py
+++ b/tests/test_aio.py
@@ -4,6 +4,7 @@ import asyncio
 # import uvloop
 import threading
 import random
+from unittest.mock import patch
 
 # asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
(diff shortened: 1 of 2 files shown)


aisk:
Maybe a little test case on this should be cool?

aiudirog:
@aisk I mocked out TAsyncSocket and validated that the arguments are handled as expected. I also squashed the commits since I had some small hiccups with flake8 and Py3.5.

aisk:
Great job 👍 LGTM

aiudirog:
I'll go ahead and merge this in. Thanks!

ethe:
👍 good job

Issue #122, opened 2020-03-03
typo correction：Grammer should be "Grammar"


truebit:
https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/parser/exc.py#L14

PR #123, opened 2020-03-03
Fix typo word

truebit:
fix issue #122

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -207,7 +207,7 @@ def test_e_dead_include():
 def test_e_grammer_error_at_eof():
     with pytest.raises(ThriftGrammerError) as excinfo:
         load('parser-cases/e_grammer_error_at_eof.thrift')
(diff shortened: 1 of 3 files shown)


aisk:
I think we should have an alias for `ThriftGrammerError` for don't break unit test and users' codes.

And maybe we can declare a module level `__getattr__` which descibed in [PEP 562](https://www.python.org/dev/peps/pep-0562/), for users who were using Python3.7 and latter versions can got a deprecated warning.

truebit:
IMHO, this change do not need to consider backward compatibility due to the ROI of this change. Anyone used old `ThriftGrammerError` would be easy to find out the changed class name.

ethe:
I think it would be better to keep backward compatibility, I have to release a minor version after this pull request be merged, it is not necessary

aisk:
The only cocern is the error message is changed, but I think this will only affect thriftpy2's test.

ethe (approved):

PR #128, opened 2020-03-10
Fix exception handling in TProcessor

microdog:
This fixes the issue that exceptions may be dropped silently in TProcessor. The issue was introduced in https://github.com/Thriftpy/thriftpy2/pull/4.


ethe:
LGTM

PR #129, opened 2020-03-16
Support Cython in HTTP and fix TCyBufferedTransport early flush issue

aiudirog:
**TL;DR**
This PR solves two issues:
- TCyBufferedTransport flusing the underlying transport before the message is complete
- HTTP client only supporting pure Python TBufferedTransport

-----

While developing an alternative HTTP client to the one in thriftpy2.http using HTTPX (so I can use the same GSSAPI authentication on both sync and async) I ran into the issue [described here:](https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/http.py#L68)

>  Also, using TCyBufferedTransportFactory will let THttpClient write a broken string to server, which making server freezed in transport.readall() method.

After some debugging, I discovered that this is because TCyBufferedTransport [flushes its underlying transport when its buffer size will be exceeded by a write.](https://github.com/Thriftpy/thriftpy2/tree/b78720e22905ba48e044b7a0d1437def3aec923d/thriftpy2/transport/buffered/cybuffered.pyx#L54) For streaming transports, like TSocket, this isn't an issue. However, for HTTP (and other message based transports) the flush method creates a request and sends that to the server. The request is supposed to be a complete Thrift message, however the early flushing behavior of TCyBufferedTransport causes an incomplete message to be sent. 

My proposed solution is to not flush the underlying transport until `TCyBufferedTransport`'s `flush()` or `c_flush()` methods are called directly. I did this by adding `c_dump_wbuf()` which checks if there is data to write and, if so will, writes the data into the underlying transport and then clears the write buffer. Now, when the write buffer is too full, this method is used to empty the buffer into the underlying transport without flushing it. `c_flush()` can then be simplified to dump the write buffer and then flush the underlying transport.

------------
For anyone else who comes across the other issues listed in the comments of http.py:
- `Cannot convert TBufferedTransport to thriftpy2.transport.cybase.CyTransportBase` is caused by [TCyBinaryProtocol declaring the transport as a CyTransportBase for performance.](https://github.com/Thriftpy/thriftpy2/tree/b78720e22905ba48e044b7a0d1437def3aec923d/thriftpy2/protocol/cybin/cybin.pyx#L410) Wrapping your custom transport in TBufferedTransport (which would end up being TCyBufferedTransport) solves this issue because it doesn't care about the type of the wrapped transport.
- `too small buffer allocated by TCyBufferedTransport` is caused by [passing a buffer size less than 1024 to TCyBufferedTransport.](https://github.com/Thriftpy/thriftpy2/tree/b78720e22905ba48e044b7a0d1437def3aec923d/thriftpy2/transport/buffered/cybuffered.pyx#L20) The HTTP client is passing this value explicitly to make sure only one call was every made to `self.rfile.read(size)` because an over-read like the buffered transports normally do can cause the read socket to timeout. I'm not sure why this occurs, since normal sockets typically just return what they have at the moment up to size, but I fought with it for about an hour before deciding to work around it.

I was able to fix these two issues and enable arbitrary server transport support by simply pre-reading the full content into a BytesIO (which is what was internally happening anyways with the buf_size hack) and passing that to whatever transport factory the user chooses.

### DIFF
--- a/thriftpy2/http.py
+++ b/thriftpy2/http.py
@@ -217,13 +217,16 @@ class THttpClient(object):
         self.__wbuf.write(buf)
 
     def flush(self):
+        # Pull data out of buffer
+        # Do this before opening a new connection in case there isn't data
+        data = self.__wbuf.getvalue()
+        self.__wbuf = BytesIO()
+        if not data:  # No data to flush, ignore
+            return
+
         if self.isOpen():
             self.close()
         self.open()
-
-        # Pull data out of buffer
-        data = self.__wbuf.getvalue()
-        self.__wbuf = BytesIO()
 
         # HTTP request
         self.__http.putrequest('POST', self.path, skip_host=True)
--- a/thriftpy2/http.py
+++ b/thriftpy2/http.py
@@ -54,23 +54,10 @@ else:
 
 from thriftpy2.thrift import TProcessor, TClient
 from thriftpy2.server import TServer
-from thriftpy2.transport import (
-    TTransportBase,
-    TMemoryBuffer
-)
-# Explicitly use Python version instead of Cython version for libraries below
-# to address some mystery issues for now.
-#
-# Avoid TypeError: Cannot convert TBufferedTransport to
-# thriftpy2.transport.cybase.CyTransportBase.
-from thriftpy2.protocol.binary import TBinaryProtocolFactory
-# Avoid raised error of too small buffer allocated by TCyBufferedTransport.
-# Also, using TCyBufferedTransportFactory will let THttpClient write a broken
-# string to server, which making server freezed in transport.readall() method.
-from thriftpy2.transport.buffered import (
-    TBufferedTransport,
-    TBufferedTransportFactory,
-)
+from thriftpy2.transport import TTransportBase, TMemoryBuffer
+
+from thriftpy2.protocol import TBinaryProtocolFactory
+from thriftpy2.transport import TBufferedTransportFactory
 
 
 HTTP_URI = '{scheme}://{host}:{port}{path}'
@@ -120,6 +107,7 @@ class THttpServer(TServer):
     def __init__(self,
                  processor,
                  server_address,
+                 itrans_factory,
                  iprot_factory,
                  server_class=http_server.HTTPServer):
         """Set up protocol factories and HTTP server.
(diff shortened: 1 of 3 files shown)


aiudirog:
I tested 3.6->3.8 before pushing, looks like I need to be more thorough. I'll figure out what's up with 2.7 and 3.5 before tomorrow.

ethe:
I find that you modify the behavior of `TCyBufferedTransport.c_write`, would it also change the behavior of other servers rather than HTTP only?

aiudirog:
I changed it to just not flush the underlying transport, which is how the regular TBufferedTransport works already. I went through every transport and only TCyBufferedTransport calls flush on the wrapped transport inside its write, so I would say that it is safe to say that it is in the wrong and anything relying on that behavior shouldn't have been, especially with flush being a top level call.

The only tests which are failing are the HTTP ones and that is for a different reason altogether I'm debugging now.

ethe:
I agree.

ethe:
I checked the pure Python version of TBufferedTransport, you are right.

aiudirog:
Okay, I found the bug: [`TCyBinaryProtocol` calls `flush()` in `write_message_end()`](https://github.com/Thriftpy/thriftpy2/blob/b78720e22905ba48e044b7a0d1437def3aec923d/thriftpy2/protocol/cybin/cybin.pyx#L463) but no other protocol does. [`TClient._send()` flushes the transport directly right after calling `write_message_end()`](https://github.com/Thriftpy/thriftpy2/blob/b78720e22905ba48e044b7a0d1437def3aec923d/thriftpy2/thrift.py#L228) which means there is a double flush when using `TCyBinaryProtocol` compared to the other protocol implementations. So, the HTTP client sends an empty request because it send a request every flush.

I see this as two bugs: 
1. The HTTP client should ignore a double flush when it doesn't have data 
2. The binary protocol shouldn't be flushing in the message end since no other protocol does

Thoughts?

ethe:
I think we should correct the `TCyBinaryProtocol` as other protocols.

aiudirog:
Awesome, I'll push my commit and let's see if it works. I tested on 2.7 this time, so fingers crossed.

aiudirog:
FYI, I noticed that the coverage reports aren't running, even in master. Might be something to check out.

aiudirog:
:) I'll merge it in then, thanks for being available so quickly on a Monday morning.

ethe:
There is a duplicated job in CI, code coverage works fine in https://travis-ci.com/github/Thriftpy/thriftpy2/jobs/298409113. I will try to remove another one.

aiudirog:
Ah got it, no problem then.

aiudirog:
@ethe Do you think you can do a release within the next week or so? 

By the way, here is the client I'm developing: https://github.com/aiudirog/ThriftPy2-HTTPX-Client. I decided not to send it as a PR to contrib since it's 3.6+ only and adds additional dependencies.

ethe:
OK, I will launch a release this week.

PR #132, opened 2020-05-28
thrift: treat required fields as positional arguments while generating init, closes #119

### LINKED ISSUE #119: struct required field is not verified
Similar as #72 , required field in struct should be enforced before send out. Current implementation only validate method args.

Since the init func is generated on the fly, so maybe we should add some validation code in __init__?

```python

    """Generate `__init__` function based on TPayload.default_spec

    For example::

        spec = [('name', 'Alice'), ('number', None)]

    will generate a types.FunctionType object representing::

        def __init__(self, name='Alice', number=None):
            self.name = name
            self.number = number

            if number is None: raise ValueError()
```

This way actually breaks init then assign behavior. Or add a validate method on each generated class and call it just before send?

iamsudip:


thriftpy2/thrift.py
```diff
@@ -68,9 +68,22 @@ def __init__(self):
             pass
         return __init__
 
+
     varnames, defaults = zip(*spec)
```
iamsudip, line 72: @ethe This function still returning the same function as before, just that init generated would have positional arguments now cause I didn't change `defaults` variable which is being passed to `types.FunctionType` while returning. What do you think?

iamsudip: Unless we change defaults, I don't think this commit makes any difference.

```
In [4]: ab = thriftpy2.load("tests/base.thrift")

In [5]: ab.Hello.__init__
Out[5]: <unbound method Hello.__init__>

In [6]: ab.Hello.__init__??
Signature: ab.Hello.__init__(self, greet=None, name=None)
Source:   
def __init__(self, greet, name=None):
    self.name = name
File:      ~/Work/Personal/thriftpy2/<generated Hello.__init__>
Type:      instancemethod

In [9]: ab.Hello()
Out[9]: Hello(name=None, greet=None)
```
This should raise error, right?

ethe: Yes, I think raising error here should be better, maybe we can remove required arguments from defaults.

thriftpy2/thrift.py
```diff
@@ -68,9 +68,22 @@ def __init__(self):
             pass
         return __init__
 
+
```
ethe, line 71: It is better to keep single line here.

thriftpy2/thrift.py
iamsudip: Removing required arguments from defaults breaking lots of tests. Look at new change and breaking test cases. I guess we can keep the signature intact and add `ValueError` statements inside __init__ if args are None and if required.

iamsudip: @ethe Look at this commit: https://github.com/Thriftpy/thriftpy2/pull/132/commits/5617a5520686c767f96848a89c23704c3f599a04

thriftpy2/thrift.py
```diff
@@ -68,9 +68,26 @@ def __init__(self):
...
+        kwargs = []
+        varnames = []
+        defaults = []
+        for spec_element, t_spec in zip(spec, cls.thrift_spec.values()):
+            varnames.append(spec_element[0])
+            if t_spec[-1]:
+                args.append(spec_element[0])
+            else:
+                kwargs.append(spec_element)
+                defaults.append(spec_element[1])
```
iamsudip, line 82: This is breaking lot's of valid tests

ethe: I checked that it is all crashed in Client._send function, I think it is easy to be fixed.

iamsudip: @ethe Can you tell me how can I achieve that?

I tried changing args_to_kwargs to not make args to kwargs, passed them to _send, but later failed at `TProcessor:process_in:args = getattr(self._service, api + "_args")()` because of required args and we removed them from defaults.

iamsudip: Or we can add a fixture to be added to generated __init__ for raising ValueError if required args don't have value.

ethe: Compare with raising ValueError in runtime, I think it is better to use args because of a friendly function signature. I will fetch your branch and maybe research for a while. Thank you!

iamsudip: @ethe Sure. I also think the same that using `args` is the way to go, I faced blocker here: `Processor` dependency on generated service api. Would love to learn from your implementation.

Issue #136, opened 2020-09-05
mixed up binary and string types


vitrun:
why does thriftpy2 treat `binary` as `string`?

in [thrift.py](https://github.com/Thriftpy/thriftpy2/blob/90cbc37191364a82ec8a847c5c578024defcca83/thriftpy2/thrift.py#L88):

```python
    STRING = 11
    UTF7 = 11
    BINARY = 11  # This here just for parsing. For all purposes, it's a string
```

this may cause inconvenience when sending files, images, etc.

ethe:
There is no difference between string and binaries in Python 2, so it is not be considered before.

ethe:
Close via #139

Issue #13, opened 2018-10-23
Question about JSON protocol


wreed4:
I'm trying to use the TJSONProtocolFactory with the client I'm generating.  And from what I can tell, the JSON it is generating is not valid thrift json.  For example for a given function call on the client, I'm getting something like this: 

```json
'{"metadata": {"version": 1, "name": "registerThing", "ttype": 1, "seqid": 0}, 
"payload": {"request": {"uuid": "12322312121", "url": "someUrl", "user": "JohnSmith", "password": "pass1234"}}}'
```
In contrast I think I would be expecting something closer to this (the last object part is almost definitely wrong, but that kind of format anyways):
```json
[1, "registerThing, 1, 0, { "1": {"rec" : {"1": {"str": "12322312121"}, "2": {"str": "someUrl"...............
```
or something like that.  

I am very new to Thrift, so I could be wrong here, but I'm confused as to how I'm supposed to use this Protocol.  The server I'm trying to connect to is rejecting the request as thriftpy sends it (unfortunately I can't give too much detail on that as it is a proprietary system).  I noticed this difference in the json by debugging the request from the client I'm trying to write and comparing it to the requests received from other clients written in other languages.

ethe:
Could you please post your thrift IDL file here? I will try to reproduce it.

wreed4:
Unfortunately, I can't paste the thrift file I'm using as it is proprietary.  However, here is a very paired-down version of the specific endpoint/type i'm trying to use.

```thrift
struct Foo {
  1: required string uuid
  2: required string url
  3: required string user
  4: required string password
}

struct FooRequest {
  1: required string uuid
  2: required string url
  3: required string user
  4: required string password
}

struct FooResponse {
  1: required Foo foo
}


service Api {
  FooResponse doFoo(1: FooRequest request)
}

```

The json I get back is 
```json
'{"metadata": {"version": 1, "name": "doFoo", "ttype": 1, "seqid": 0}, 
"payload": {"request": {"uuid": "12322312121", "url": "adjdafslkdfkljfad", "user": "23jk", "password": "j23"}}}'
```

This does not look like the format I expect for thrift.  
I'm trying to use TMemoryBuffer (wrote my own factory) and TJSONProtocol.

ethe:
[...]

PR #139, opened 2020-10-01
Added Support For Apache JSON and Binary Data

JonnoFTW:
#13 Added support for apache json, so that apache thrift server/clients can talk to thriftpy2 server/clients.
#136 Added support for binary data fields, I'm not sure what ramifications this would have on python2.


ethe:
Some CI has failed, do you mind to fix it?

JonnoFTW:
@ethe don't merge yet, I have some problems in my own codebase where `TApacheJSONProtocol` does not work with `THttpClient`

JonnoFTW:
@ethe feel free to check it out now

thriftpy2/__init__.py
```diff
@@ -5,7 +5,7 @@
 from .hook import install_import_hook, remove_import_hook
 from .parser import load, load_module, load_fp
 
-__version__ = '0.4.11'
+__version__ = '0.4.12'
```
ethe, line 8: Please do not modify the version, I will upgrade the version when the package releasing.

thriftpy2/protocol/__init__.py
```diff
@@ -23,5 +24,6 @@
 __all__ = ['TBinaryProtocol', 'TBinaryProtocolFactory',
            'TCyBinaryProtocol', 'TCyBinaryProtocolFactory',
            'TJSONProtocol', 'TJSONProtocolFactory',
+           'TApacheJSONProtocol', 'TApacheJSONProtocolFactory',
            'TMultiplexedProtocol', 'TMultiplexedProtocolFactory',
-           'TCompactProtocol', 'TCompactProtocolFactory']
+           'TCompactProtocol', 'TCompactProtocolFactory', ]
```
ethe, line 29: Maybe it is better to leave it as before.

thriftpy2/thrift.py
```diff
@@ -104,6 +104,7 @@ class TType(object):
     LIST = 15
     UTF8 = 16
     UTF16 = 17
+    BINARY = 18
```
ethe, line 107: There was already a BINARY member at https://github.com/Thriftpy/thriftpy2/pull/139/files#diff-b55308698cb9187e83350298d1d941b56d1e956b1e36f36c00fe61e68145cef5R100

thriftpy2/protocol/__init__.py
JonnoFTW: Fixed

thriftpy2/thrift.py
JonnoFTW: Fixed in 6d1b8c9

thriftpy2/__init__.py
JonnoFTW: Fixed in 8bb4ed6

thriftpy2/thrift.py
ethe: Does it affect other protocols like binary? Maybe we should check it in all protocols.

ethe: Other protocols can not treat the new BINARY member if the value of it is not `11` anymore.

ethe:
Good job and thank you, if we can confirm that it does not change features which already exist, I think we can merge it right now.

JonnoFTW:
@ethe I updated asyncio protocols to support binary fields and so far all the tests work. Might be a need to do some interoperability tests with other implementations.

ethe (approved):

setup.py
```diff
@@ -15,6 +15,7 @@
 
 install_requires = [
     "ply>=3.4,<4.0",
+    "six>=1.15.0"
```
aisk, line 18: I thought we should pin the max version of six, for six 2.0 may have breaking changes. `six~=1.15` is suitable here.
and it's better to have comma and the end of line.

aisk (changes requested):

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+        suitable_for_isinstance = (suitable_for_isinstance,)
+    for thing in suitable_for_isinstance:
+        if isinstance(thing, tuple):
+            types.extend(flatten(thing))
+        else:
+            types.append(thing)
+    return tuple(types)
+
+
+class TApacheJSONProtocolFactory:
```
aisk, line 56: All classes should be inherit from object for we have python2 compatible.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+        return TApacheJSONProtocol(trans)
+
+
+class TApacheJSONProtocol(TProtocolBase):
+    """
+    Protocol that implements the Apache JSON Protocol
+    """
+
+    def __init__(self, trans):
+        TProtocolBase.__init__(self, trans)
```
aisk, line 68: It's better to call `super(TApacheJSONProtocol, self).__init__(trans)`.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+        pass
+
+    def write_message_begin(self, name, ttype, seqid):
+        self.api = name
+        self.ttype = ttype
+        self.seqid = seqid
+
+    def write_struct(self, obj):
+        """
+        Write json to self.trans following apache style jsonification of `obj`
```
aisk, line 117: Missing new line between comments and type annotations.

thriftpy2/protocol/compact.py
```diff
@@ -108,7 +110,8 @@ class CompactType(object):
     TType.STRUCT: CompactType.STRUCT,
     TType.LIST: CompactType.LIST,
     TType.SET: CompactType.SET,
-    TType.MAP: CompactType.MAP
+    TType.MAP: CompactType.MAP,
+    TType.BINARY: CompactType.BINARY
```
aisk, line 114: It's better to have a comman here.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def read_struct(self, obj):
+        """
+        Read the next struct into obj, usually the argument from an incoming request
+        Only really used to read the arguments off a request into whatever we want
+        see thriftpy2.thrift.TProcessor.process_in for how this class will be used
+
+        Will turn the contents of self.req[4] into the args of obj,
+        ie. self.req[4]["1"] must be rendered into obj.thrift_spec
+        :param obj:
```
aisk, line 277: Same as mentioned above.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+                else:
+                    result[field_idx] = {
+                        CTYPES[ttype]: val
+                    }
+        return result
+
+    def _dict_to_thrift(self, data, base_type):
+        """
+        Convert an apache thrift dict (where key is the type, value is the data)
+        :param data:
```
aisk, line 197: Same as mentioned above.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
+# -*- coding: utf-8 -*-
+"""
+Transport for json protocol that apache thrift files will understand
+unfortunately, thriftpy2's TJSONProtocol is not compatible with apache's
+"""
+from __future__ import absolute_import
```
aisk, line 6: Missing newline here.

thriftpy2/protocol/apache_json.py
```diff
@@ -0,0 +1,280 @@
...
+import json
+import base64
+
+from six import string_types
```
aisk, line 10: Missing newline here.

tests/test_apache_json.py
```diff
@@ -0,0 +1,171 @@
+from __future__ import absolute_import
```
aisk, line 1: Missing coding magic line here.

thriftpy2/protocol/apache_json.py
JonnoFTW: I copied the usage from existing protocol classes:

https://github.com/Thriftpy/thriftpy2/blob/17a66b5c3fdef6dd66c5d853c08eafcbffb0b7e4/thriftpy2/protocol/binary.py#L363

https://github.com/Thriftpy/thriftpy2/blob/17a66b5c3fdef6dd66c5d853c08eafcbffb0b7e4/thriftpy2/protocol/compact.py#L127

https://github.com/Thriftpy/thriftpy2/blob/17a66b5c3fdef6dd66c5d853c08eafcbffb0b7e4/thriftpy2/protocol/json.py#L166

Unless you want to change all of them to `super` calls?

setup.py
JonnoFTW: Fixed in d87d5cc

tests/test_apache_json.py
JonnoFTW: Fixed in ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/apache_json.py
JonnoFTW: Fixed ffb624c

thriftpy2/protocol/compact.py
JonnoFTW: Fixed 23ab765

setup.py
```diff
@@ -15,7 +15,7 @@
 
 install_requires = [
     "ply>=3.4,<4.0",
-    "six>=1.15.0"
+    "six~=1.15.0",
```
aisk, line 18: It should be `six~=1.15`, the `six~=1.15.0` have different meanings. See here: https://www.python.org/dev/peps/pep-0440/#examples

thriftpy2/protocol/apache_json.py
aisk: I think we should not modify the old codes in this PR. Maybe we can change it in the new codes only?

JonnoFTW: @aisk Feel free to create an issue and I will fix it in a 2nd pull request.

aisk:
> Hi @JonnoFTW I have a question, is the changes for #136 will make a incompatible change for current users for both python2 & 3? If so, I think we should and a flag to enable this feature, for we don't breaking current users' codes in minor changes.

Hi @JonnoFTW what about this question?

thriftpy2/thrift.py
JonnoFTW: Other protocols have been updated to support the  `BINARY` type

JonnoFTW:
@aisk tests pass for python2 and python3 so I don't think it breaks and existing code since the old tests still pass with the assumption that they have coverage for `STRING` and `BINARY` thrift types.

ethe (approved):

aisk:
> @aisk tests pass for python2 and python3 so I don't think it breaks and existing code since the old tests still pass with the assumption that they have coverage for `STRING` and `BINARY` thrift types.

Really sorry for the late of reply. I tried to understand the changes for #136 and took a dig and found there may have some breaking changes.

For example, before this PR, if we have a thrift file that returns a binary field, and it's content is valid utf-8 contents like "hello", the user will got an unicode in Python2 (str in Python3), but after this PR, he will got str in Python (bytes in Python3). If the content only contains ASCII, maybe we can use str / unicode mixed up, but for other cases this will brings trouble.

I think maybe our test case does not covered this case, if you have any doubt about this, maybe I can add some tests for this.

But for now, I think it's cool if you can split this PR into two, the changes for #13 looks good to me.

JonnoFTW:
I've created a small test repo to check that my changes don't break backwards compatibility with Apache Thrift's python implementation (and hopefully other versions too).

https://github.com/JonnoFTW/thrift_interop_test/blob/master/tests/test_interop.py

thriftpy2/protocol/binary.py
```diff
@@ -225,6 +228,10 @@ def read_val(inbuf, ttype, spec=None, decode_response=True):
     elif ttype == TType.DOUBLE:
         return unpack_double(inbuf.read(8))
 
+    elif ttype == TType.BINARY:
+        sz = unpack_i32(inbuf.read(4))
+        return inbuf.read(sz)
```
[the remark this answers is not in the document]
JonnoFTW: @aisk what do you want me to do then?

I've checked in the apache version and they return `bytes` for a binary type field.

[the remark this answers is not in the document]
aisk: I mean thriftpy2 compatibility, previous vedrsion byted field may return unicode.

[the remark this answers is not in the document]
JonnoFTW: I ran the following test in python 2.7.18:

```py
from __future__ import print_function
import os
import time
from multiprocessing import Process

import thriftpy2
from thriftpy2.rpc import make_client, make_server

cdir = os.path.dirname(os.path.realpath(__file__))
test_thrift = thriftpy2.load(cdir + "/spec.thrift")
Bar = test_thrift.Bar
Foo = test_thrift.Foo


def test_img():
    with open(cdir + "/test.png", 'rb') as infile:
        data = infile.read()
    obj = Bar(
        tbinary=data,
        tlist_of_binary=[b'a binary string']
    )

    class Handler(object):
        def test(self, t):
            return t

    def do_server():
        server = make_server(service=test_thrift.BarService, handler=Handler(), host='localhost', port=9090)
        server.serve()

    proc = Process(target=do_server)
    proc.start()
    time.sleep(0.1)

    client = make_client(test_thrift.BarService, host='localhost', port=9090)
    res = client.test(obj)
    print("Type of binary=", type(res.tbinary))
    print("Type of t_list_of_binary=", type(res.tlist_of_binary[0]))
    proc.terminate()


if __name__ == "__main__":
    test_img()

```

With `thriftpy2==0.4.10` I get:

```
Type of binary= <type 'str'>
Type of t_list_of_binary= <type 'unicode'>
```

With `thriftpy2==0.4.12` I get:
```
Type of binary= <type 'str'>
Type of t_list_of_binary= <type 'str'>
```

Is this going to break anyone's code?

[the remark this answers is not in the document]
aisk: Hi @JonnoFTW , I tried to run your codes on Python2, got the same output with master branch. But the got the output with current PR:

```
Type of binary= <type 'NoneType'>
Type of t_list_of_binary= <type 'NoneType'>
```

But I think maybe you are right, the previous behavior seems more like a bug and nobody should rely on it. What's your opinion @ethe ?

aisk (approved):

aisk:
Hi @JonnoFTW I approved this PR, thanks!

JonnoFTW:
@ethe are you happy to merge now? All tests pass along with those in my interoperability repo.

ethe:
Merged, thank you for your patience and contribution.

thriftpy2/protocol/binary.py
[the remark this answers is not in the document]
ethe: @aisk I think it does not break the backward compatibility.

dmulter:
I've been using Thriftpy2 with a private Thrift server for a long time now, and this latest addition has broken things for me. Sorry I can't share the Thrift files, but here's the error when making an RPC call:

```
.virtualenv/lib/python3.8/site-packages/thriftpy2/thrift.py:219: in _req
    return self._recv(_api)
.virtualenv/lib/python3.8/site-packages/thriftpy2/thrift.py:231: in _recv
    fname, mtype, rseqid = self._iprot.read_message_begin()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   cybin.ProtocolError: No protocol version header

thriftpy2/protocol/cybin/cybin.pyx:454: ProtocolError
```

JonnoFTW:
@dmulter are you able to provide a test case to recreate this? Also did you install from pypi or from git?

JonnoFTW:
Can you try installing from git?

dmulter:
Fails the same way.

JonnoFTW:
@dmulter can you please give a minimum working example or just tell me exactly what combination of client/server protocols and transports you are using on each end. I'll add it to the tests and see if I can recreate the problem.

ethe:
@dmulter could you please provide a sample IDL file? we need to reproduce it.

JonnoFTW:
@dmulter I'm unable to replicate the problem you are having, here's my test case (it still passes if I use `TCompactProtocol`):

Spec:

```
struct BinarySpec {
    1: map<string, binary> str2bin,
    2: map<string, string> str2str,
}

service BinaryService {
    BinarySpec run(1: BinarySpec b_spec);
}
```

```python
def test_tcompact_memory_buffer():
    spec = load("binary_test.thrift")

    factory = cybin.TCyBinaryProtocolFactory()
    obj = spec.BinarySpec(
        str2bin={"key": b"Binary Value"},
        str2str={"key": "String Value"}
    )

    class Handler:
        def run(self, b_spec):
            print("\nServer received:")
            print(b_spec)
            return b_spec

    def run_server():
        server = make_server(
            service=spec.BinaryService,
            handler=Handler(),
            proto_factory=factory,
        )
        server.serve()

    proc = Process(target=run_server)
    try:
        proc.start()
        time.sleep(0.5)
        client = make_client(spec.BinaryService, proto_factory=factory)
        res = client.run(obj)
        assert res
        print("Server sent:")
        print(res)
    finally:
        proc.kill()
    time.sleep(0.2)
```

Gives the output:

```
Server received:
BinarySpec(str2bin={'key': b'Binary Value'}, str2str={'key': 'String Value'})
Server sent:
BinarySpec(str2bin={'key': b'Binary Value'}, str2str={'key': 'String Value'})
```


If you could modify this test so that it fails in the same way your code does, that would be very useful in fixing the issue.

dmulter:
Some additional status:

* I thought I had installed from a Git clone, but just tried as part of my testing and I get the following error during build: `clang: error: no such file or directory: 'thriftpy2/transport/cybase.c'`
* I verified that the test you created above works fine. I had to change things a bit to avoid multiprocessing errors that are also included in the next bullet item code.
* I created a version that makes the specific service call and it also works fine. See code at the end.
* Seems clear to me that the issue requires the actual server implementation since local tests are not exposing the problem. My strategy now is trying to identify why the Thriftpy2-0.4.13 release broke when talking to the server. I'm trying to build a local version that I can use for debugging the returned bytes to help identify what the code is choking on. I first have to solve the first item above. Note that I can't give access to the real server since it is not ours and is both private and protected via a layer of complex authentication that must be performed first. I experimented with various other transports and was not able to find a way to make the test pass.

xxx.thrift:
```python
enum ApiVersion
{
    kApiVersion1 = 1;
    kApiVersion2;
    kApiVersion3;
}
const ApiVersion kServerApiVersion_Current = ApiVersion.kApiVersion2;

service PrivateService
{
    void openSession
        (
        1: string sessionId,
        2: ApiVersion clientVersion
        )
}
```

test_xxx.py:
```python
import time
import sys
from thriftpy2 import load
from thriftpy2.rpc import make_client, make_server
from multiprocessing import Process

class Handler:
    def openSession(self, sessionId, clientVersion):
        print(f"\nServer received: sessionId={sessionId} clientVersion={clientVersion}")
        sys.stdout.flush()

def run_server():
    print("Starting server...")
    server = make_server(
        service=spec.PrivateService,
        handler=Handler(),
    )
    server.serve()

spec = load("xxx.thrift")

def test_private_service():
    proc = Process(target=run_server)
    try:
        proc.start()
        time.sleep(0.5)

        print("Calling openSession...")
        client = make_client(service=spec.PrivateService)
        session_id = "xxx"
        client.openSession(session_id, spec.kServerApiVersion_Current)
        assert False, "tests actually worked" # so we can see print output
    finally:
        proc.kill()
    time.sleep(0.2)
```

dmulter:
Installed Cython and built from source. Currently experimenting with what's returned from the server. Not sure if it helps, but the `size` value it expects for the protocol version header has a value of `3c68746d`. Trying to get the full server response...

dmulter:
Found the problem and it was not this PR, but a different change on 0.4.13. The protocol error was because HTML for an error page was being returned. The breaking change was due to an inserted path separator after port as I'm passing a parsed URL as follows:

```python
    from urllib.parse import urlparse

    url = urlparse(full_url)
    client = make_client(
        service=spec.TestService,
        scheme=url.scheme,
        host=url.netloc,
        path=url.path,
        port="443"
    )
```

And the change was https://github.com/Thriftpy/thriftpy2/pull/148/commits/faf10551188fc23e3293c526e1142abbdf2a9d14

JonnoFTW:
@ethe considering that the @dmulter 's issue didn't come from this pull request, can we merge it back in? The slash problem can be fixed separately.

ethe:
@JonnoFTW yeah your merge does not break compact protocol, however, there is another problem, see #156

PR #125, opened 2020-03-03
Support include files with dot in name

aisk:
Not a completed work, for now just support "one level include", like including a file called "com.example.api.thrift".

On the opposite, including a file called "com.example.thrift", which also including another file "api.thrift", is not supported now.

So this PR is marked as WIP, I'll try to finish this latter.

Close #121

PR #141, opened 2020-10-09
fix multi level module

windard:
Change-Id: I59002abfd3d2dd5665ddee8a6a12d1513f15758a

### DIFF
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -403,7 +403,7 @@ def p_ref_type(p):
                 ref_type = resolved_ref_type
                 break
     else:
-        for index, name in enumerate(p[1].split('.')):
+        for index, name in enumerate(p[1].rsplit('.', 1)):
             ref_type = getattr(ref_type, name, None)
             if ref_type is None:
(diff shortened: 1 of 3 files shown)


aisk:
@windard do you have any related issue for this PR? And I think it should be cool if there was a test case for we don't break the behavior in this PR in the future.

thriftpy2/parser/parser.py
```diff
@@ -403,7 +403,7 @@ def p_ref_type(p):
                 ref_type = resolved_ref_type
                 break
     else:
-        for index, name in enumerate(p[1].split('.')):
+        for index, name in enumerate(p[1].rsplit('.', 1)):
```
aisk, line 406: The split count is set to one, so I thought this will not work with thrift files with multiple dot in name.

aisk: And why change the split to rsplit? I think the resolve order is chaned and I thought it's a ad-hoc to your case and will not work with other kind of cases.

windard:
yeah, it has been fixed, but not released. let's close this PR.
https://github.com/Thriftpy/thriftpy2/pull/125

PR #148, opened 2020-11-12
Fix HTTP_URI typo in http.py

### LINKED ISSUE #140: Invalid HTTP URI format in http.py
https://github.com/Thriftpy/thriftpy2/blob/a8b7873078a6f38a2f052964d439ab888e32fcf4/thriftpy2/http.py#L63

It's missing the `/`  between the port and the path and should be:

```python
HTTP_URI = '{scheme}://{host}:{port}/{path}'
```

This only occurs if you leave out the `/` in your path variable.

aisk:
close #140

### DIFF
--- a/thriftpy2/http.py
+++ b/thriftpy2/http.py
@@ -60,7 +60,7 @@ from thriftpy2.protocol import TBinaryProtocolFactory
 from thriftpy2.transport import TBufferedTransportFactory
 
 
-HTTP_URI = '{scheme}://{host}:{port}{path}'
+HTTP_URI = '{scheme}://{host}:{port}/{path}'
 DEFAULT_HTTP_CLIENT_TIMEOUT_MS = 30000  # 30 seconds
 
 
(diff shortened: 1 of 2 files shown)


ethe (approved):

thriftpy2/http.py
```diff
@@ -60,7 +60,7 @@
 from thriftpy2.transport import TBufferedTransportFactory
 
 
-HTTP_URI = '{scheme}://{host}:{port}{path}'
+HTTP_URI = '{scheme}://{host}:{port}/{path}'
```
dmulter, line 63: This change just broke things for me.

aisk:
Really sorry for the break, so should we add a if statement and and the `/` if it not exists?

dmulter:
Your call, but since the standard in `urlparse` is to have a starting slash, I would think it's the responsibility of the caller to make sure the path is in that format. Could also be clarified in the docs. Anyway, that would be my opinion.

aisk:
@dmulter I think you are right, and really sorry for I did not make it right. Just open a new PR: https://github.com/Thriftpy/thriftpy2/pull/158 , can you help me review it?

PR #149, opened 2020-11-12
add support for custom headers for HTTP

ermeaney:
Our thrift implementation needs to have some custom headers in the HTTP call, and since the implementation was almost there I updated the make_client function to support it

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -6,15 +6,16 @@ import os
 import multiprocessing
 import socket
 import time
+import uuid
 
 import pytest
 
 import thriftpy2
+
 thriftpy2.install_import_hook()  # noqa
 
-from thriftpy2.http import make_server, make_client, client_context  # noqa
+from thriftpy2.http import make_server, make_client, client_context, THttpHeaderFactory  # noqa
 from thriftpy2.thrift import TApplicationException  # noqa
-
 
 addressbook = thriftpy2.load(os.path.join(os.path.dirname(__file__),
                                           "addressbook.thrift"))
@@ -64,6 +65,11 @@ class Dispatcher():
     def sleep(self, ms):
         time.sleep(ms / 1000.0)
         return True
+
+
+class CustomHeaderFactory(THttpHeaderFactory):
+    def get_headers(self):
+        return {"X-REQUEST-ID": str(uuid.uuid4())}
 
 
 @pytest.fixture(scope="module")
@@ -121,6 +127,35 @@ def client_without_url(timeout=3000):
                        port=6080, path="foo", timeout=timeout)
 
 
+def client_context_with_header_factory(timeout=3000):
+    return client_context(addressbook.AddressBookService,
+                          url="http://127.0.0.1:6080",
+                          timeout=timeout,
+                          http_header_factory=THttpHeaderFactory())
+
+
+def client_context_with_custom_header_factory(timeout=3000):
+    return client_context(addressbook.AddressBookService,
+                          url="http://127.0.0.1:6080",
+                          timeout=timeout,
+                          http_header_factory=CustomHeaderFactory())
+
+
+def client_with_header_factory(timeout=3000):
+    return make_client(addressbook.AddressBookService,
+                       url="http://127.0.0.1:6080",
+                       timeout=timeout,
+                       http_header_factory=THttpHeaderFactory(
+                           {"X-REQUEST-ID": str(uuid.uuid4())}))
+
+
+def client_with_custom_header_factory(timeout=3000):
+    return make_client(addressbook.AddressBookService,
+                       url="http://127.0.0.1:6080",
+                       timeout=timeout,
+                       http_header_factory=CustomHeaderFactory())
+
+
 def test_client_context(server):
     with client() as c1, client_context_with_url() as c2:
         assert c1.hello("world") == c2.hello("world")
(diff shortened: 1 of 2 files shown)


ethe:
Hi, please add a test case to this feature, thanks.

thriftpy2/http.py
```diff
@@ -282,6 +282,7 @@ def make_client(service, host='localhost', port=9090, path='', scheme='http',
                 proto_factory=TBinaryProtocolFactory(),
                 trans_factory=TBufferedTransportFactory(),
                 ssl_context_factory=None,
+                custom_headers=None,
```
aisk, line 285: I think a `http_headers_factory` is more suitable since we can specify headers per request.

ermeaney: It took me a minute to figure out what you were referring to, but I got it now and am working through that. After thinking about it I realized that we actually need exactly what you are referring to.

I am trying to figure out how to implement test cases around this feature, but am struggling a little since it looks like the HTTP request is sort of hidden in the API, so if you have some pointers that would be great.

I will update the PR once I get an implementation put together.

ermeaney: @aisk I have updated this PR to provide a header factory as suggested
@ethe I updated test_http.py to add test cases for the different ways that the factory can be used

Let me know if there is anything else

thriftpy2/http.py
```diff
@@ -159,7 +177,7 @@ class THttpClient(object):
     """Http implementation of TTransport base.
     """
 
-    def __init__(self, uri, timeout=None, ssl_context_factory=None):
+    def __init__(self, uri, timeout=None, ssl_context_factory=None, http_header_factory=THttpHeaderFactory()):
```
aisk, line 180: Default value in a keyword arguments is actually static (global) variable, which will leads to misleading usage. I think it's better to use None as the default value, and check if it's None and initialize it in function body.

thriftpy2/http.py
```diff
@@ -99,6 +99,24 @@ def __init__(self, handler):
...
+        @param headers(dict) a dictionary of static headers the factory generates
+        """
+        self.__headers = headers
+
+
+    def __enter__(self):
+        return self
+
+
+    def get_headers(self):
```
aisk, line 116: I think it's better to ensure that `THttpHeaderFactory` will return a dict instead of None.

thriftpy2/http.py
```diff
@@ -99,6 +99,24 @@ def __init__(self, handler):
...
+    """Default header factory return no custom headers
+    """
+    def __init__(self, headers=None):
+        """Initialize a header factory
+        @param headers(dict) a dictionary of static headers the factory generates
+        """
+        self.__headers = headers
+
+
+    def __enter__(self):
```
aisk, line 112: Why there is a enter function here?

thriftpy2/http.py
aisk: Sorry @ermeaney I missed the comment here ... The codes you updated in this PR is just the same as I think, I think it's ok, just a few small comments here.

thriftpy2/http.py
ermeaney: That is on me, I was having problems when using a context and thought this was the fix, turns out I was wrong. Updated

thriftpy2/http.py
ermeaney: Fixed in latest update

thriftpy2/http.py
ermeaney: Fixed in latest update

thriftpy2/http.py
ermeaney: @aisk I resolved the things you pointed out in the comments, and thanks for the pointers

ethe (approved):

tests/test_http.py
```diff
@@ -116,6 +122,35 @@ def client_with_url(timeout=3000):
...
+                       timeout=timeout,
+                       http_header_factory=THttpHeaderFactory(
+                           {"X-REQUEST-ID": str(uuid.uuid4())}))
+
+
+def client_with_custom_header_factory(timeout=3000):
+    return make_client(addressbook.AddressBookService,
+                       url="http://127.0.0.1:6080",
+                       timeout=timeout,
+                       http_header_factory=CustomHeaderFactory)
```
aisk, line 151: Looks like we are missing a parenthese here.

tests/test_http.py
```diff
@@ -133,6 +168,28 @@ def test_clients(server):
...
+
+
+def test_client_with_header_factory(server):
+    c = client_with_header_factory()
+    assert c.hello("world") == "hello world"
+    c.close()
+
+
+def test_client_with_custom_header_factory(server):
+    c = client_with_header_factory()
```
aisk, line 188: I think you want call `client_with_custom_header_factory` here, and this typo caused my previous comment's error not exposed.

thriftpy2/http.py
```diff
@@ -178,7 +194,10 @@ def __init__(self, uri, timeout=None, ssl_context_factory=None):
             self.path += '?%s' % parsed.query
         self.__wbuf = BytesIO()
         self.__http = None
-        self.__custom_headers = None
+        if http_header_factory:
```
aisk, line 197: `self._http_header_factory = http_header_factory or THttpHeaderFactory()` is more readable here.

thriftpy2/http.py
```diff
@@ -235,18 +254,17 @@ def flush(self):
         self.__http.putheader('Host', self.host)
         self.__http.putheader('Content-Type', 'application/x-thrift')
         self.__http.putheader('Content-Length', str(len(data)))
-
-        if (not self.__custom_headers or
-                'User-Agent' not in self.__custom_headers):
+        if (not self._http_header_factory.get_headers() or
```
aisk, line 257: the `self._http_header_factory.get_headers()` may have side effect, like someone's custom header factory may add a count on it, or have a performance concern like reading config from database.

So I think we should call it just once in this function and store it in a variable.

ermeaney: that was a great catch, thanks/

I just pushed an update with fixes

thriftpy2/http.py
ermeaney: done

tests/test_http.py
ermeaney: you are correct, fixed

tests/test_http.py
ermeaney: fixed

aisk (approved):

aisk:
LGTM, approved, thanks for the contribution @ermeaney !

But I saw there was a conflict, I think you should rebase the commits so we can merge it.

PR #151, opened 2020-11-25
Build a wheel as part of the Make build step so that it will be published to pypi

dbaxa:


dbaxa:
cc @aisk

ethe (approved):

ethe:
Approved, thank you.

dbaxa:
@ethe how do we merge this :-) ?

aisk:
I think maybe we have future works to do, since thriftpy2 have C extensions, the `bdist_wheel` only produce wheel package for current python version * CPU arch * OS. I think we need a CI cluster to build the wheels.

PR #158, opened 2021-01-26
Fix the path issue caused by #148

aisk:
Bofore #148, `path` in `make_client` / `client_context` should have `/` prefix. After #148 been merged, `path` must have no `/` prefix, this will break exists codes using it, and as @dmulter said, `path` should have `/`.

In this PR, thriftpy2 will support these two usecases both, and the no `/` prefix is malformed but I think we should not break exists code again.


dmulter:
Looking forward to a new release that includes this fix!

ethe:
I will release it as soon as possible, thanks

PR #160, opened 2021-01-28
WIP: translate socket.timeout to TTransportException

aisk:
So we can get better error message for which ip / port is called.


aisk:
Seems the test cases are excepecting excepting `socket.timeout`, so this will be a breaking change?

PR #163, opened 2021-02-09
Added Apache thrift support

JonnoFTW:
per #156 I have separated them into two branches. This branch implements apache's json protocol in  thriftpy2

Issue #156, opened 2021-01-21
v0.4.13 serialization break


ethe:
relay to #139, a quick verification:
```python
from thriftpy2.transport.memory import TCyMemoryBuffer
from thriftpy2.protocol import cybin as proto
from thriftpy2.thrift import TType

b1 = TCyMemoryBuffer()
proto.write_val(b1, TType.MAP, {"hello": "1"},
                spec=(TType.STRING, TType.STRING))
b1.flush()

b2 = TCyMemoryBuffer()
proto.write_val(b2, TType.MAP, {"hello": b"1"},
                spec=(TType.STRING, TType.BINARY))
b2.flush()

b1.getvalue() != b2.getvalue()
```

TType was be specified in Apache Thrift standard, and Binary type should always have the same behavior with String, hence we missed some check to modify the value of Binary: 18 to String: 11 in complex type serialization (map, list, etc.).

ethe:
@JonnoFTW @aisk @dmulter

ethe:
I think maybe I have to revert this merge #139

ethe:
Holding this issue before Apache JSON protocol re-merge.

JonnoFTW:
I'll have a look at this when I get back to work on Monday

ethe:
@JonnoFTW Thanks for your work! I have several suggestions:
- could you split support the Binary type feature and Apache JSON protocol in two pull requests?
- we should ensure that the Binary type has the same behavior with String outside of JSON-like protocols
- the Binary type maybe should allow the user to pass Python `str` in with an implicit type converting?

JonnoFTW:
@ethe 

> * could you split support the Binary type feature and Apache JSON protocol to two pull requests?

How do you want Thrifpy2's Apache JSON to handle the case where it doesn't support binary? Apache thriftpy will send binary data as base64 encoded string. If we don't support binary in Apache JSON, then the result will be that thriftpy2 gets a base64 encoded string. From the test, without binary support, apache json will make the following field:

```py
tbinary=b"\x01\x0fabc123\x00\x02"
```

Into this:

```py
{'str': 'AQ9hYmMxMjMAAg=='}
```

Without binary support, my apache thrift protocol has no way of knowing whether or not to decode the above base64 encoded string.

> * we should ensure that the Binary type has the same behavior with String outside of JSON-like protocols

What do you mean? Please provide an example.

> * the Binary type maybe should allow the user to pass Python `str` in with an implicit type converting?

[...]

PR #164, opened 2021-02-10
Binary support

JonnoFTW:
Added binary support per #156 

I've added some more tests to cover more cases surrounding binary encodings. If you pass a string to a binary field, it will come out as a bytes type on the other end.

The work here builds off #163 , so that will probably need to be merged before this one.


JonnoFTW:
@ethe are you able to look at this at some point?

ethe:
I will take a look at it, could you please remove [WIP] in the title?

JonnoFTW:
@ethe I have fixed the title.

ethe (approved):

ethe:
I have a question, does it break the compatibility? If the user defines a binary type and treats it as a string before, what happened after this commit merged?

ethe:
Maybe we should consider that only turn on this feature in JSON protocol, and do not modify others, or turn off it as default.

JonnoFTW:
@ethe please check `tests/test_all_protocols_binary_field.py` for a test case on how binary fields are treated. I think the best behaviour would be that the type specified in the thrift file is what you get on the when deserializing.

If a string is passed to a binary field, when deserialised you always get a binary. Conversely, if a binary is passed to a string field, it should be stringified  (if possible) when deserializing.

PR #175, opened 2021-05-17
fix issue that py3.4 cannot use pytest due to latest attr dropping support.

### LINKED ISSUE #174: ImportError: No module named 'typing' on python 3.4 due to attr drop support on >=21.1.0
See this job: https://travis-ci.com/github/Thriftpy/thriftpy2/jobs/505479622
And attr issue: https://github.com/python-attrs/attrs/pull/807

truebit:
try to fix #174

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,8 @@ commands =
     py.test
 
 deps =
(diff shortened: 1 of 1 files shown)


ethe:
LGTM, thanks

PR #178, opened 2021-07-16
Fix error not raised when using a not defined type

### LINKED ISSUE #177: error not raised when type from another file is missing
cat b.thrift:

```thrift
enum Status {
    OK = 1
    ERROR = 2
}
```

cat a.thrift:

```thrift
include "b.thrift"

struct A {
    1: b.Status status
    2: b.InvalidType type  // notice that this type is not defined in b.
}
```

parse the files and dump the `thrift_spec` for struct A:

```python
import thriftpy2


a = thriftpy2.load("a.thrift")
print(a.A.thrift_spec)
```

And got the result:

```python
{1: (8, 'status', <class 'b.Status'>, False), 2: (<module 'b'>, 'type', False)}
```

As we can see, the field with number 2 have incomplete type, and there is no error been raised, and until someone called a function which returns a `struct A`, we got a missleading error message:

```
TypeError: an integer is required
```

aisk:
close #177

### DIFF
--- a/tests/parser-cases/issue_177.thrift
+++ b/tests/parser-cases/issue_177.thrift
@@ -1,0 +1,6 @@
+include "issue_177_include.thrift"
+
+struct Issue177 {
+    1: issue_177_include.Status status
+    2: issue_177_include.invalidType type
+}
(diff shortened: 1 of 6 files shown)


ethe:
LGTM

PR #180, opened 2021-08-04
yield from is necessary when using coroutine in aio

lexdene:
Some functions returns a coroutine.
`yield from` is necessary for them.


ethe:
if you you are using async client / server, yes

lexdene:
May I ask when this patch will be published to pypi?

PR #181, opened 2021-08-05
close socket after client timeout

lexdene:
I found the socket is not closed correctly after `asyncio.wait_for` timeout.
So add `writer.close()` in `TimeoutError` exception.


lexdene:
@ethe please take a look.

aisk (approved):

PR #183, opened 2021-12-17
close TSocket socket after server disconnect

pawl:
This is similar to: https://github.com/redis/redis-py/pull/1797

Currently, `TSocket.close()` won't continue to close the socket if it encounters an `OSError` on `socket.shutdown`. An `OSError` can happen on `socket.shutdown` if the server side of the connection disconnected already. Not running `socket.close` prevents garbage collection from cleaning up the socket and will use more memory than necessary (maybe even causing a leak?).

This PR puts a separate try/except around `socket.close` and `socket.shutdown` to allow the socket to be closed after a failed shutdown.

### DIFF
--- a/tests/test_socket.py
+++ b/tests/test_socket.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division
 import os
 import socket
 import sys
+from unittest import mock
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


ethe:
LGTM

PR #184, opened 2021-12-20
close TServerSocket socket after server disconnect

pawl:
This is the same as this other PR that was recently merged: https://github.com/Thriftpy/thriftpy2/pull/183

But it fixes the issue in `TServerSocket` too.


aisk (approved):

pawl:
@ethe would you be up for merging this one? It's pretty much the same as this other one you merged recently: https://github.com/Thriftpy/thriftpy2/pull/183

PR #202, opened 2022-11-14
Fixed not specified size issue in read_val(aio)

chozo99:
There was a size in the existing code, 
but it seems to have been removed while applying async/await.

reference: <https://github.com/Thriftpy/thriftpy2/commit/b4abbdd14dbedbb7f057021d222d7d5b9631ecd1>


ethe (approved):

ethe:
thanks!

PR #211, opened 2023-03-29
fix: support annotations in const

LYF1999:
fix: support annotations in const


ethe (approved):

ethe:
LGTM, cc @aisk

ethe:
> Can you add some tests for this case?

I think `tests/parser-cases/annotations.thrift` is the test case of supporting annotation of constant value. Same as other kinds of annotation, we just need to parse it and skip all passes behind.

aisk (approved):

PR #222, opened 2023-09-01
Add sasl transport support

ecederstrand:
Adds pure Python and Cython implementation of sasl transport.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,6 @@ import platform
 from os.path import join, dirname
 
 from setuptools import setup, find_packages
-from setuptools.command.build_ext import build_ext
 from setuptools.extension import Extension
 
 with open(join(dirname(__file__), 'thriftpy2', '__init__.py'), 'r') as f:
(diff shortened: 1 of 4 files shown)


ethe:
Thanks, I am reviewing it.

setup.py
```diff
@@ -39,30 +40,32 @@
...
+        Extension("thriftpy2.transport.cybase", sources=["thriftpy2/transport/cybase.pyx"]),
+        Extension("thriftpy2.transport.buffered.cybuffered", sources=["thriftpy2/transport/buffered/cybuffered.pyx"]),
+        Extension("thriftpy2.transport.framed.cyframed", sources=["thriftpy2/transport/framed/cyframed.pyx"]),
+        Extension("thriftpy2.transport.memory.cymemory", sources=["thriftpy2/transport/memory/cymemory.pyx"]),
+        Extension("thriftpy2.transport.sasl.cysasl", sources=["thriftpy2/transport/sasl/cysasl.pyx"]),
+        Extension("thriftpy2.protocol.cybin", sources=["thriftpy2/protocol/cybin/cybin.pyx"]),
+    ]
+
+
+class CustomBuildExtCommand(build_ext):
```
aisk, line 59: Hi, I want to know that's the proposal for change the setup process?

The pip / setuptools have changed so much now days, and we specified the cython dependency in `pyproject.toml`. So in the `pip install` (or other modern python package mangagement tools), the cython will be installed automatically. So I think the `build_ext` command is not needed.

thriftpy2/transport/sasl/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+""" SASL transports for Thrift. """
+
+# Initially copied from the Impala repo
```
aisk, line 20: I think we should add a URL for the original file which is copied from.

aisk:
I think this change need some test cases to ship with. And since this PR only implment the sasl transport client, maybe we can call some public thrift services with sasl support?

setup.py
ecederstrand: My issue was that I was running `setup.py bdist_wheel` (I believe, will have to double-check) in an empty venv on a Unix system. That broke because setup.py expects Cython to be pre-installed. By adding the `CustomBuildExtCommand`, setuptools has time to install Cython as a build requirement before building the extension.

thriftpy2/transport/sasl/__init__.py
ecederstrand: Done :-9

ecederstrand:
I agree that tests would be useful. I don't know of any public thrift services with sasl support, so would really like any suggestions here, or alternative methods for testing this code. We have internal an HBase installation that we tested this against, but a full HBase is quite a task to build for a test setup.

setup.py
aisk: For local development, use `pip install -e .` in a venv (or modern tool like rye / pipenv). The `python setup.py ...` way is to be eliminated.

aisk:
Fully implment the sasl server transport is the best way. 

Another way is starting a TCP server before running this test case, and the server can be a dummpy server which only send a valid thrift response package while accept and read fixed length of packet.

You can dump a thrift packat via wireshard or something like it.

setup.py
ecederstrand: I just tested with `python3 -m build --wheel` on the main branch and that works without my changes.

I still think that adding a `CustomBuildExtCommand` is the "correct" way to create extension modules, but I can revert the patch to keep your approach and only add the new sasl transport source file.

aisk: I think this is not related to the sasl transport implementation, so I suggest just revert it, and maybe we can discuss about it in a separate thread.

aisk:
Since this is a new feature and it won't break current users' code, I think we can merge this without testing, if @ecederstrand doesn't have time to finish the work.

@ethe any concerns?

ethe:
> Since this is a new feature and it won't break current users' code, I think we can merge this without testing, if @ecederstrand doesn't have time to finish the work.

It looks good to me, maybe we could publish a pre-release version that contains this experimental feature.

aisk (approved):

PR #235, opened 2024-02-25
Fix deprecated usage of pytest

aisk:
Details: 

- https://docs.pytest.org/en/stable/deprecations.html#setup-teardown
- https://docs.pytest.org/en/stable/deprecations.html#using-pytest-warns-none

About the change in the `test_rpc.py`, see the comment below.

## Summary by CodeRabbit

- **Tests**
	- Standardized method names for setup and teardown operations in test classes.
	- Removed a deprecated warning in the `test_client_timeout` function related to client instantiation.

### DIFF
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -250,9 +250,8 @@ def test_exception_iwth_ssl():
 
 def test_client_timeout():
     with pytest.raises(socket.timeout):
-        with pytest.warns(UserWarning):  # Deprecated
-            with client(timeout=500) as c:
-                c.sleep(1000)
+        with client(timeout=500) as c:
+            c.sleep(1000)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_rpc.py
```diff
@@ -250,9 +250,8 @@ def test_exception_iwth_ssl():
 
 def test_client_timeout():
     with pytest.raises(socket.timeout):
-        with pytest.warns(UserWarning):  # Deprecated
```
aisk, line 253: This `UserWarning` is never been raised, but the older version of pytest will ignore the failure when it't in a `with pytest.raises(...)` context.

So we should just remove it.

aisk:
Since this is a trival change which keeps the CI healthy, I'll merge it after the CI passed. cc @ethe

PR #249, opened 2024-03-08
Add submodule to sys.path when loading child idl file.

cocolato:
Fixes: https://github.com/Thriftpy/thriftpy2/issues/215

I'm not sure if it's suitable to add the module directly to `sys.path`, maybe is should implement like this:
```python
for module in thrift.__thrift_meta__["includes"]:
    if module not in sys.modules:
        sys.modules[module.__name__+"_thrift"] = module
```
 then use the import hook to make sure pickle will import the suitable module.

### DIFF
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -39,6 +39,9 @@ def load(path,
         fill_incomplete_ttype(thrift, thrift)
     if real_module:
         sys.modules[module_name] = thrift
+        for module in thrift.__thrift_meta__["includes"]:
+            if module not in sys.modules:
+                sys.modules[module.__name__] = module
     return thrift
 
 
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -37,11 +37,16 @@ def load(path,
                    include_dir=include_dir, encoding=encoding)
     if incomplete_type:
         fill_incomplete_ttype(thrift, thrift)
+
+    # add sub modules to sys.modules recursively
     if real_module:
         sys.modules[module_name] = thrift
-        for module in thrift.__thrift_meta__["includes"]:
+        sub_modules = thrift.__thrift_meta__["includes"][:]
+        while sub_modules:
+            module = sub_modules.pop()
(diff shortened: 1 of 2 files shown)


ethe (approved):

thriftpy2/parser/__init__.py
```diff
@@ -39,6 +39,9 @@ def load(path,
         fill_incomplete_ttype(thrift, thrift)
     if real_module:
         sys.modules[module_name] = thrift
+        for module in thrift.__thrift_meta__["includes"]:
```
aisk, line 42: Do we need add these modules recursively?

cocolato: Sure, if the current solution if what we want，there should rewrite to recursion.

cocolato:
Now has been updated to a recursive implementation.

thriftpy2/parser/__init__.py
```diff
@@ -37,8 +37,16 @@ def load(path,
...
+
+    # add sub modules to sys.modules recursively
     if real_module:
         sys.modules[module_name] = thrift
+        sub_modules = thrift.__thrift_meta__["includes"][:]
+        while sub_modules:
+            module = sub_modules.pop()
+            if module not in sys.modules:
+                sys.modules[module.__name__] = module
+                sub_modules.extend(module.__thrift_meta__["includes"])
```
aisk, line 49: Should we copy the `includes` like L44?

```suggestion
                sub_modules.extend(module.__thrift_meta__["includes"][:])
```

aisk:
I have some concern about whether this code can handle circle including IDLs. Can you add a test case to ensure this?

cocolato:
Ok, i'll test this part later.

cocolato:
> I have some concern about whether this code can handle circle including IDLs. Can you add a test case to ensure this?

There are already have tests for dead inclusions.
https://github.com/Thriftpy/thriftpy2/blob/91b28ea3de4bc3d734364d9e902b8a2ddee5fc80/tests/test_parser.py#L201-L204

aisk (approved):

PR #252, opened 2024-03-20
Remove python2/python3.5 historical compatibility code

cocolato:


thriftpy2/_compat.py
```diff
@@ -11,11 +11,9 @@
 
 import platform
 import sys
-from urllib.parse import urlparse
-from urllib.request import urlopen
+from urllib.parse import urlparse  # noqa
+from urllib.request import urlopen  # noqa
```
aisk, line 15: If no codes are using these modules, I think we should remove them.

```suggestion
```

thriftpy2/rpc.py
```diff
@@ -7,7 +7,8 @@
 import urllib
 import warnings
 
-from thriftpy2._compat import PY35
+from thriftpy2.contrib.aio.rpc import make_client as make_aio_client  # noqa
+from thriftpy2.contrib.aio.rpc import make_server as make_aio_server  # noqa
```
aisk, line 11: These two lines can be removed too.

```suggestion
```

thriftpy2/_compat.py
cocolato: https://github.com/Thriftpy/thriftpy2/blob/61ef458059c63fb1fd0181780123de1f36462825/thriftpy2/parser/parser.py#L17 it will import urlparse  from _compat,  I think we should import from urllib directly here and maybe should remove it in another pr

thriftpy2/rpc.py
cocolato: https://github.com/Thriftpy/thriftpy2/blob/61ef458059c63fb1fd0181780123de1f36462825/README.rst#L95-L101
https://github.com/Thriftpy/thriftpy2/blob/61ef458059c63fb1fd0181780123de1f36462825/examples/asyncio_echo/server.py#L2-L5
The ducument and example rely on this.

ethe (approved):

PR #257, opened 2024-04-03
Run test on pypy

aisk:


tests/test_all_protocols_binary_field.py
```diff
@@ -263,12 +260,12 @@ def run_server():
     time.sleep(0.2)
 
 
-@pytest.mark.skipif(cybin is None, reason="Must be run in cpython")
+@pytest.mark.skipif(_compat.PYPY, reason="Must be run in cpython")
```
aisk, line 263: Because we are missing a `__init__.py` in the `cybin` folder, modern version of Python (not Python2) will treat this folder as a namespace module, so it's not `None`.

tests/test_protocol_cybinary.py
```diff
@@ -11,15 +11,16 @@
...
-
-pytestmark = pytest.mark.skipif(PYPY,
-                                reason="cython not enabled in pypy.")
 if not PYPY:
     from thriftpy2.protocol import cybin as proto
     from thriftpy2.transport.buffered import TCyBufferedTransport
     from thriftpy2.transport.memory import TCyMemoryBuffer
 
 
+if PYPY:
```
aisk, line 20:
```suggestion
else:
```

PR #258, opened 2024-04-06
Support cythonized ext on Windows

aisk:
I found that it's very easy to support it on Windows. The only issue in the current master branch is that `endian_port.h` doesn't support Windows.

So I tried to use https://gist.github.com/panzi/6856583 to replace the old one. It's also used by Google's Fuchsia: https://fuchsia.googlesource.com/third_party/iperf/+/3.1.5/src/portable_endian.h


aisk:
Needs rebase with #258 before landing.

Issue #204, opened 2022-11-21
thriftpy2 whether support using in ipv6


mengyaommm:
thriftpy2 seems to support ipv6, but there is no configuration parameter in Entrance function
![image](https://user-images.githubusercontent.com/40377342/202971370-54a67ff1-638b-456a-bad9-5bc9ff1351ba.png)
![image](https://user-images.githubusercontent.com/40377342/202971337-98d9f652-9530-4a52-bbe6-2143cb56bf4b.png)

PR #261, opened 2024-04-15
Support using ipv6 in make_client/make_server method

cocolato:
Fix: https://github.com/Thriftpy/thriftpy2/issues/204

### DIFF
--- a/thriftpy2/contrib/aio/rpc.py
+++ b/thriftpy2/contrib/aio/rpc.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import socket
 import urllib
 import warnings
 
@@ -17,7 +18,7 @@ async def make_client(service, host='localhost', port=9090, unix_socket=None,
                       cafile=None, ssl_context=None,
                       certfile=None, keyfile=None,
                       validate=True, url='',
-                      socket_timeout=None):
+                      socket_timeout=None, socket_family=socket.AF_INET):
     if socket_timeout is not None:
         warnings.warn(
             "The 'socket_timeout' argument is deprecated. "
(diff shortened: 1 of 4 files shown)


ethe (approved):

thriftpy2/contrib/aio/rpc.py
```diff
@@ -56,18 +59,22 @@ def make_server(service, handler,
...
-                keyfile=None, ssl_context=None, loop=None):
+                keyfile=None, ssl_context=None, loop=None,
+                socket_family=socket.AF_INET):
     processor = TAsyncProcessor(service, handler)
 
     if unix_socket:
-        server_socket = TAsyncServerSocket(unix_socket=unix_socket)
+        server_socket = TAsyncServerSocket(
+            unix_socket=unix_socket,
+            socket_family=socket_family)
```
aisk, line 69: If I remember correctly, `AF_INET` is the same level as `AF_UNIX`, which means we shouldn't specify the `socket_family` to `AF_INET` for unix domain socket.

aisk (changes requested):

thriftpy2/contrib/aio/rpc.py
cocolato: Thanks, updated.

aisk:
There are still some codes where `socket_family` is applied to unix domain sockets.

cocolato:
> There are still some codes where `socket_family` is applied to unix domain sockets.

Does this mean that when users pass in the `unix_socket` option, they will not be supported to pass in the additional `socket_family` option?

aisk:
Yes, the unix domain socket's `socket_family` must be `AF_UNIX`. See: https://www.ibm.com/docs/en/i/7.3?topic=characteristics-socket-address-family

thriftpy2/contrib/aio/rpc.py
```diff
@@ -30,22 +31,34 @@ async def make_client(service, host='localhost', port=9090, unix_socket=None,
...
-        socket = TAsyncSocket(unix_socket=unix_socket,
-                              connect_timeout=connect_timeout,
-                              socket_timeout=timeout)
+        socket_family = socket.AF_UNIX if not socket_family else socket_family
+        client_socket = TAsyncSocket(
+            unix_socket=unix_socket,
+            connect_timeout=connect_timeout,
+            socket_timeout=timeout,
+            socket_family=socket_family,
+        )
```
cocolato, line 40: Ok, I think I should not pass the socket_family option when using unix_socket.

```suggestion
        client_socket = TAsyncSocket(
            unix_socket=unix_socket,
            connect_timeout=connect_timeout,
            socket_timeout=timeout
        )
```

cocolato: I will rewrite this part later today.

aisk: Exactly

cocolato:
> Yes, the unix domain socket's `socket_family` must be `AF_UNIX`. See: https://www.ibm.com/docs/en/i/7.3?topic=characteristics-socket-address-family

Updated.

PR #263, opened 2024-04-18
Support use mp spawn mode for tests

cocolato:


cocolato:
Maybe 3.6 testing should be skipped. I'm not sure if this implementation is what we want.

aisk:
I think this approach is fine for me. But there are some tests that have been skipped because they require `fork` to start multiprocessing. So with this change, if this approach can solve the problem, we should enable these skipped tests to ensure that this change works as we expected.

For skipping the test on Windows with Python 3.6, I think maybe `multiprocess` does not support some functionalities on Windows with Python 3.6. So I'm fine with skipping it, but since it passed on Linux, I think we can modify the `build.yaml` to have it only skip on Windows with Python 3.6.

Issue #232, opened 2024-01-29
Thrift IDL file parse reports error in multi-threaded situation


1C4nfaN:
When my idl definition is more complex, I import idl files in multiple threads at the same time, and some exceptions will be thrown during the parser process.

Looking at the code, I found that the parser process does not seem to be thread-safe. If I want the code to work properly, I Add thread locks to my thread environment. I think a documentation tip may be needed here.

aisk:
Fixed in #264

PR #264, opened 2024-05-07
Move some global variables in parser into thread local

aisk:
Related to #232.  
  
I'm not sure if this can totally resolve the original issue, because reproducing a multi-thread bug is hard. I'd like to ship the change because it wouldn't introduce any trouble, and wait to see if there will be more multi-thread issues with the parser. Otherwise, we should introduce a big lock for all these `load` methods.

### DIFF
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -13,7 +13,7 @@ import os
 import sys
 import types
 
-from .parser import parse, parse_fp, incomplete_type, _cast
+from .parser import parse, parse_fp, threadlocal, _cast
 from .exc import ThriftParserError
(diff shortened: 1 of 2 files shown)


ethe (approved):

aisk:
The failed tests are caused by https://github.com/pytest-dev/pytest/issues/12263, as that issue said it will be fixed on tornado or pytest repository, so I'll leave as it is for now.

truebit:
This change would invoke error `AttributeError: '_thread._local' obiect has no attribute 'thrift_stack' ` when using it with example code:
```
>>> from thriftpy2.parser. parser import parse
>>> note_thrift = parse("path/ to/ note. thrift")
```

Issue #270, opened 2024-05-28
[BUG]  Title: Asynchronous Client Hangs When Connecting to Non-Existent IP


fuergaosi233:
**Description:**
When using an asynchronous client to connect to a non-existent IP, the program hangs indefinitely because the socket operation is pending globally. This issue causes the entire program to be stuck.

**Steps to Reproduce:**
1. Use the following code to create an asynchronous client that attempts to connect to a non-existent IP.
2. The program will hang and not proceed beyond the client connection attempt.

**Code to Reproduce:**
```python
# -*- coding: utf-8 -*-
import thriftpy2
import asyncio
from thriftpy2.rpc import make_aio_client

echo_thrift = thriftpy2.load("echo.thrift", module_name="echo_thrift")
import datetime

async def print_current_time():
    while True:
        print(datetime.datetime.now())
        await asyncio.sleep(1)

async def main():
    task_current_time = asyncio.create_task(print_current_time())
    await asyncio.sleep(5)
    client = await make_aio_client(
        echo_thrift.EchoService, "100.100.100.100", 6000, connect_timeout=100* 1000
    )
    print(await client.echo("hello, world"))
    client.close()

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

**Expected Behavior:**
The connection attempt should timeout after the specified `connect_timeout` duration and raise an appropriate exception, allowing the program to handle it gracefully and don't block loop.

**Actual Behavior:**
The program hangs indefinitely and does not raise an exception after the timeout period. This behavior prevents the rest of the program from executing.

**Environment:**
- `thriftpy2` version: 0.5.1rc1
- Python version: 3.9
- Operating System: Debian GNU/Linux

**Additional Information:**
The issue appears to be related to how the socket operation is handled when connecting to an unreachable IP address. This results in the entire program being stuck, which can be problematic for applications relying on asynchronous operations.

PR #271, opened 2024-05-28
🐛 Fix async pending error

fuergaosi233:
fix https://github.com/Thriftpy/thriftpy2/issues/270

### DIFF
--- a/thriftpy2/contrib/aio/socket.py
+++ b/thriftpy2/contrib/aio/socket.py
@@ -64,6 +64,8 @@ class TAsyncSocket(object):
             to persist SSLContext object. Caution it's easy to get wrong, only
             use if you know what you're doing.
         """
+        self.read_sock = None
+        self.write_sock = None
         if sock:
             self.raw_sock = sock
         elif unix_socket:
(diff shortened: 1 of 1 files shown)


aisk:
Please avoid formatting the code you are not touching.

thriftpy2/contrib/aio/socket.py
```diff
@@ -106,58 +105,23 @@ def __init__(self, host=None, port=None, unix_socket=None,
...
-        linger = struct.pack('ii', 0, 0)
-        _sock.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, linger)
-        _sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
-
-        self.raw_sock = _sock
-
-    def set_handle(self, sock):
-        self.raw_sock = sock
-
-    def set_timeout(self, ms):
```
aisk, line 126: I think we still can implement this method, just set the instance fields, which can be used in open and read methods.

thriftpy2/contrib/aio/socket.py
```diff
@@ -106,19 +107,34 @@ def __init__(self, host=None, port=None, unix_socket=None,
...
-    def _init_sock(self):
-        if self.unix_socket:
-            _sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        else:
-            _sock = socket.socket(self.socket_family, socket.SOCK_STREAM)
-            _sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+    async def open(self):
+        addr = self.unix_socket or (self.host, self.port)
+        try:
+            if self.unix_socket:
```
ethe, line 113: use self.socket_factory instead, see https://github.com/Thriftpy/thriftpy2/pull/271/files#diff-19812d639b869b73a1158c80d996c6fa56787f93905759034caac0381ed7f8fdR82

PR #295, opened 2024-11-04
fix: sub module conflict error

StellarisW:
Currently the logic of register sub module to `sys.modules` may cause conflict,

That's because of the PR https://github.com/Thriftpy/thriftpy2/pull/249 which fix https://github.com/Thriftpy/thriftpy2/issues/215

```py
def load(path,
         module_name=None,
         include_dirs=None,
         include_dir=None,
         encoding='utf-8'):

    real_module = bool(module_name)
    thrift = parse(path, module_name, include_dirs=include_dirs,
                   include_dir=include_dir, encoding=encoding)
    if threadlocal.incomplete_type:
        fill_incomplete_ttype(thrift, thrift)

    if real_module:
        sys.modules[module_name] = thrift
        sub_modules = thrift.__thrift_meta__["includes"][:]
        while sub_modules:
            module = sub_modules.pop()
            if module not in sys.modules:
                sys.modules[module.__name__] = module
                sub_modules.extend(module.__thrift_meta__["includes"])
    return thrift
```

For example:

We have thrift files as follows:
```
thrift files
├── idl
│   ├── main.thrift
│   ├── included1.thrift
│   ├── included2.thrift
```

If we call `load('idl/main.thrift', module_name='idl.main_thrift')`, then the following modules will be registered:

- `idl.main_thrift`
- `included1_thrift`
- `included2_thrift`

So, if `included1.thrift` is a common module name like [platform](https://docs.python.org/3/library/platform.html), which is `platform.thrift`, then module `platform` will register to `sys.modules`, so it will cause conflict.

Now I come up with 2 solutions:

1. Add an option to `load()`, which will be `def load(path, module_name=None, include_dirs=None, include_dir=None, encoding='utf-8')`, and let the users to register sub modules manually. **(not recommended)**

2. Optimize sub module name. **(recommended)**

   use absolute path as its module name instead of the base name of the thrift file.

    let's take the example before, then the following modules will be registered:

   - `idl.main_thrift`
   - `idl.included1_thrift`
   - `idl.included2_thrift`

### DIFF
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -14,7 +14,7 @@ import sys
 import types
 
 from .parser import parse, parse_fp, threadlocal, _cast
-from .exc import ThriftParserError
+from .exc import ThriftParserError, ThriftModuleNameConflict
 from ..thrift import TPayloadMeta
 
 
@@ -42,21 +42,21 @@ def load(path,
     if real_module:
         sys.modules[module_name] = thrift
         include_thrifts = list(zip(thrift.__thrift_meta__["includes"][:],
-                                   thrift.__thrift_meta__["sub_modules"][:]))
+                                   [module.__thrift_module_name__ for module in thrift.__thrift_meta__["includes"][:]]))
         while include_thrifts:
             include_thrift = include_thrifts.pop()
-            registered_thrift = sys.modules.get(include_thrift[1].__name__)
+            registered_thrift = sys.modules.get(include_thrift[1])
             if registered_thrift is None:
-                sys.modules[include_thrift[1].__name__] = include_thrift[0]
+                sys.modules[include_thrift[1]] = include_thrift[0]
                 if hasattr(include_thrift[0], "__thrift_meta__"):
                     include_thrifts.extend(
                         list(
                             zip(
                                 include_thrift[0].__thrift_meta__["includes"],
-                                include_thrift[0].__thrift_meta__["sub_modules"])))
+                                [module.__thrift_module_name__ for module in include_thrift[0].__thrift_meta__["includes"]])))
             else:
                 if registered_thrift.__thrift_file__ != include_thrift[0].__thrift_file__:
-                    raise ThriftParserError(
+                    raise ThriftModuleNameConflict(
                         'Module name conflict between "%s" and "%s"' %
                         (registered_thrift.__thrift_file__, include_thrift[0].__thrift_file__)
                     )
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -41,12 +41,25 @@ def load(path,
     # add sub modules to sys.modules recursively
     if real_module:
         sys.modules[module_name] = thrift
-        sub_modules = thrift.__thrift_meta__["includes"][:]
-        while sub_modules:
-            module = sub_modules.pop()
-            if module not in sys.modules:
-                sys.modules[module.__name__] = module
-                sub_modules.extend(module.__thrift_meta__["includes"])
+        include_thrifts = list(zip(thrift.__thrift_meta__["includes"][:],
+                                   thrift.__thrift_meta__["sub_modules"][:]))
+        while include_thrifts:
+            include_thrift = include_thrifts.pop()
+            registered_thrift = sys.modules.get(include_thrift[1].__name__)
+            if registered_thrift is None:
+                sys.modules[include_thrift[1].__name__] = include_thrift[0]
+                if hasattr(include_thrift[0], "__thrift_meta__"):
+                    include_thrifts.extend(
+                        list(
+                            zip(
+                                include_thrift[0].__thrift_meta__["includes"],
+                                include_thrift[0].__thrift_meta__["sub_modules"])))
+            else:
+                if registered_thrift.__thrift_file__ != include_thrift[0].__thrift_file__:
+                    raise ThriftParserError(
+                        'Module name conflict between "%s" and "%s"' %
+                        (registered_thrift.__thrift_file__, include_thrift[0].__thrift_file__)
+                    )
     return thrift
 
 
--- a/thriftpy2/parser/__init__.py
+++ b/thriftpy2/parser/__init__.py
@@ -41,24 +41,20 @@ def load(path,
     # add sub modules to sys.modules recursively
     if real_module:
         sys.modules[module_name] = thrift
-        include_thrifts = list(zip(thrift.__thrift_meta__["includes"][:],
-                                   [module.__thrift_module_name__ for module in thrift.__thrift_meta__["includes"][:]]))
+        include_thrifts = thrift.__thrift_meta__["includes"][:]
         while include_thrifts:
             include_thrift = include_thrifts.pop()
-            registered_thrift = sys.modules.get(include_thrift[1])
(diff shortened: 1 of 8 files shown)


StellarisW:
Please take a review💌
@aisk @ethe

aisk:
Sorry, I can't access my development computer these days. I will review it a few days later.

But the test case should be passed before merging.

StellarisW:
> Sorry, I can't access my development computer these days. I will review it a few days later.
> 
> But the test case should be passed before merging.

That's OK, I'll fix the test cases.

thriftpy2/parser/__init__.py
```diff
@@ -41,12 +41,15 @@ def load(path,
...
-            if module not in sys.modules:
-                sys.modules[module.__name__] = module
-                sub_modules.extend(module.__thrift_meta__["includes"])
+        include_thrifts = thrift.__thrift_meta__["includes"][:]
+        while include_thrifts:
+            include_thrift = include_thrifts.pop()
+            sub_modules = thrift.__thrift_meta__["sub_modules"][:]
+            for module in sub_modules:
+                if module not in sys.modules:
+                    sys.modules[module.__name__] = include_thrift
```
cocolato, line 50:
```suggestion
            lost_modules = [
                m for m in thrift.__thrift_meta__["sub_modules"] if m not in sys.modules
            ]
            for module in lost_modules:
                sys.modules[module.__name__] = include_thrift
```

thriftpy2/parser/parser.py
```diff
@@ -949,3 +960,8 @@ def _get_ttype(inst, default_ttype=None):
     if hasattr(inst, '__dict__') and '_ttype' in inst.__dict__:
         return inst.__dict__['_ttype']
     return default_ttype
+
+def remove_suffix(s, suffix):
```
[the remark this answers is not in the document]
StellarisW: Thanks for the advice, it seems better indeed.

thriftpy2/parser/parser.py
```diff
@@ -62,9 +62,20 @@ def p_include(p):
...
-            setattr(thrift, child.__name__, child)
+            child_path = os.path.normpath(
+                os.path.dirname(str(thrift.__name__).replace("_thrift", "").replace(".", os.sep)) + os.sep + p[2])
+
+            child_path = child_path.lstrip(os.sep)
+
+            child_module_name = str(
+                child_path).replace(os.sep,
+                                    ".").replace(
+                ".thrift", "_thrift")
```
aisk, line 73: These code's indents are a little bit strange, I think it's better to using something like:

```
aaa = bbb().ccc()\
    .ddd()
```

aisk:
Thank you for your work, but I still have a question.

Although the included Thrift modules wouldn't overwrite each other when they have the same name but are in different subdirectories after merge this PR, we support loading a Thrift file with a dot in its name, such as `foo.bar.thrift`. This means that if there is both a `foo/bar.thrift` and a `foo.bar.thrift`, they might still overwrite each other.

I think the expected behavior should be to raise an exception to inform users about the conflict. I believe the current code already has this issue, but the new code might have the potential to address it.

StellarisW:
> Thank you for your work, but I still have a question.
> 
> Although the included Thrift modules wouldn't overwrite each other when they have the same name but are in different subdirectories after merge this PR, we support loading a Thrift file with a dot in its name, such as `foo.bar.thrift`. This means that if there is both a `foo/bar.thrift` and a `foo.bar.thrift`, they might still overwrite each other.
> 
> I think the expected behavior should be to raise an exception to inform users about the conflict. I believe the current code already has this issue, but the new code might have the potential to address it.

Okay, I'll raise an exception in this case.

StellarisW:
@aisk All things have done, but there is still a test fail in `def test_load_fp()`, it because of the `load_fp()` uses `threadlocal.thrift_cache` which has been set in other test, don't know what is the best way to fix it.
CleanShot 2024-11-29 at 13 34 30@2x

cocolato:
It would be good if you could squash all the commits down to one, just to tidy up the history.

StellarisW:
> It would be good if you could squash all the commits down to one, just to tidy up the history.

done

StellarisW:
@aisk Can you please take a review, this PR is delayed quite a long time, thank you !

le119126:
I also meet the cause，waiting

cocolato:
> @aisk All things have done, but there is still a test fail in `def test_load_fp()`, it because of the `load_fp()` uses `threadlocal.thrift_cache` which has been set in other test, don't know what is the best way to fix it. CleanShot 2024-11-29 at 13 34 30@2x

try this
```python3
from thriftpy2.parser import threadlocal
threadlocal.__dict__.clear()
```

StellarisW:
> > @aisk All things have done, but there is still a test fail in `def test_load_fp()`, it because of the `load_fp()` uses `threadlocal.thrift_cache` which has been set in other test, don't know what is the best way to fix it. CleanShot 2024-11-29 at 13 34 30@2x
> 
> try this
> 
> ```python
> from thriftpy2.parser import threadlocal
> threadlocal.__dict__.clear()
> ```

The tests passed after adding the code!

> > @aisk All things have done, but there is still a test fail in `def test_load_fp()`, it because of the `load_fp()` uses `threadlocal.thrift_cache` which has been set in other test, don't know what is the best way to fix it. CleanShot 2024-11-29 at 13 34 30@2x
> 
> try this
> 
> ```python
> from thriftpy2.parser import threadlocal
> threadlocal.__dict__.clear()
> ```

Thanks, all tests have passed right now

thriftpy2/parser/parser.py
```diff
@@ -62,9 +62,23 @@ def p_include(p):
...
+            child_module_name = module_prefix + child_module_name
+
+            child = parse(path, module_name=child_module_name)
+            child_include_module_name = os.path.basename(path)
+            if child_include_module_name.endswith(".thrift"):
+                child_include_module_name = child_include_module_name[:-7]
+            child.__name__ = child_include_module_name
             setattr(thrift, child.__name__, child)
             _add_thrift_meta('includes', child)
+            _add_thrift_meta('sub_modules', types.ModuleType(child_module_name))
```
aisk, line 81: I found that this empty module is only used to check for conflicts in `sys.modules` by using it's name. Maybe we can make the code more readable by assigning the attribute `__thrift_module_name__` to the module (the variable `child` here)?

thriftpy2/parser/__init__.py
```diff
@@ -41,12 +41,25 @@ def load(path,
...
+                sys.modules[include_thrift[1].__name__] = include_thrift[0]
+                if hasattr(include_thrift[0], "__thrift_meta__"):
+                    include_thrifts.extend(
+                        list(
+                            zip(
+                                include_thrift[0].__thrift_meta__["includes"],
+                                include_thrift[0].__thrift_meta__["sub_modules"])))
+            else:
+                if registered_thrift.__thrift_file__ != include_thrift[0].__thrift_file__:
+                    raise ThriftParserError(
```
aisk, line 59: I'd like to add a new exception `ThriftModuleNameConflict` which inherits `ThriftParserError` to make it easier to catch.

StellarisW: fixed, please review the code again💌

aisk (approved):
Thanks for the work, this PR looks good to me!

ethe (approved):

Issue #187, opened 2022-03-28
Grammar error when missing field identifiers in service functions


dhuang:
I've observed that thriftpy2 throws a grammar error on load when a definition contains a service function missing its field identifiers.

Traceback with the field identifier:
```
service MyService {
  void foo(string s)
}
```

```
Traceback (most recent call last):
  File "example.py", line 2, in <module>
    thriftpy2.load('example.thrift')
  File "/path/to/thrift/env/lib/python3.7/site-packages/thriftpy2/parser/__init__.py", line 33, in load
    include_dir=include_dir)
  File "/path/to/thrift/env/lib/python3.7/site-packages/thriftpy2/parser/parser.py", line 600, in parse
    parser.parse(data)
  File "/path/to/thrift/env/lib/python3.7/site-packages/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/path/to/thrift/env/lib/python3.7/site-packages/ply/yacc.py", line 1201, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/path/to/thrift/env/lib/python3.7/site-packages/ply/yacc.py", line 192, in call_errorfunc
    r = errorfunc(token)
  File "/path/to/thrift/env/lib/python3.7/site-packages/thriftpy2/parser/parser.py", line 25, in p_error
    (p.value, p.lineno))
thriftpy2.parser.exc.ThriftGrammerError: Grammer error 'string' at line 2
```

No issue with the field identifier:
```
service MyService {
  void foo(1: string s)
}
```

It is my understand that it is best to include them, but I don't believe it's required based on the [Thrift IDL](https://thrift.apache.org/docs/idl), so this behavior is surprising.

ethe:
as you said, it is required by thrift spec to provide the backwards-compatibility of IDL

PR #304, opened 2025-04-11
Allow for omitted `FieldID` in `Field` definitions

Fluxx:
Allows `thriftpy2` to accept `Field` with optional `FieldID`.

## Background

In the Thrift IDL, [the `Field` definition](https://thrift.apache.org/docs/idl#field) looks like this:

```
[15] Field           ::=  FieldID? FieldReq? FieldType Identifier ('=' ConstValue)? XsdFieldOptions ListSeparator?
```

Of concern to this PR is the `FieldID?` portion, which indicates the field ID is optional. For instance, both of these structs are valid:

```thrift
struct Dog {
  1: string name;
}

struct Dog {
  string name;
}
```

In the first case the explicit id of `1` is given to the `name` field, while in the second case the omission of an `id` field is allowed. The correct behavior for fields which omit an id is to automatically assign them an id, starting with `-1` and decrementing backwards. As [the Thrift whitepaper](https://thrift.apache.org/static/files/thrift-20070401.pdf) says:

> To avoid conflicts between manually and automatically assigned identifiers, fields with identifiers omitted are assigned identifiers decrementing from -1, and the language only supports the manual assignment of positive identifiers.

This is what the official `thrift` compiler does, where in the generated code you see this:

```
if self.name is not None:
    oprot.writeFieldBegin('name', TType.STRING, -1)
    oprot.writeString(self.name.encode('utf-8') if sys.version_info[0] == 2 else self.name)
    oprot.writeFieldEnd()

```

## Bug

Now certainly a best practice here is to just simply not omit the `FieldID` from your IDL definitions, and the official `thrift` compiler does warn about this case.

```
[WARNING:/Users/jpollard/Code/thriftpy2/tests/parser-cases/structs.thrift:19] No field key specified for name, resulting protocol may have conflicts or not be backwards compatible!
```

But as noted above, it still generates code using the `-1` scheme. Further, fixing existing usage of the implicit `-1` scheme fields is quite tedious. Since you cannot add an explicit `-1` field id value (so as to allow `thriftpy2` to parse the file), you basically have to do a field migration, adding a new field that does explicitly set an id, migrate all users of the old field to the new one, and the remove the old field.

So ideally, `thriftpy2` would allow for a `Field` definition omitting a `FieldID`. If you try that, however, you end up with the bug reported in https://github.com/Thriftpy/thriftpy2/issues/187, which is this exception, where `token` is the field type of the field omitting its ID:

```
thriftpy2.parser.exc.ThriftGrammerError: Grammer error <token> at line 2
```

To fix this, I updated the `field_id` parser to allow for it to be omitted. If it is, it makes use of a new`threadlocal.field_seq_implicit_id` object to keep state and ensure we decrement only within the scope of the field sequence. I also added some test cases to validate the behavior.

This is my first time using a lexer and parser, and so I am not sure if I did this in a sensible way. Please feel free to let me know if there are alternative ways to approach this.

### DIFF
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -410,7 +410,7 @@ class CurrentIncompleteType(dict):
 
 
 class CurrentFieldSeqImplicitId(object):
-    def __init__(self, *args, **kwargs):
+    def __init__(self):
         self.index = -1
 
     def get_id(self):
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -8,6 +8,7 @@ IDL Ref:
 from __future__ import absolute_import
 
 import collections
+import itertools
 import os
 import threading
 import types
@@ -348,7 +349,7 @@ def p_field_seq(p):
     '''field_seq : field sep field_seq
                  | field field_seq
                  |'''
-    threadlocal.field_seq_implicit_id = CurrentFieldSeqImplicitId()
+    threadlocal.field_seq_implicit_id = itertools.count(start=-1, step=-1)
     _parse_seq(p)
 
 
@@ -379,7 +380,7 @@ def p_field_id(p):
     '''field_id : INTCONSTANT ':'
                 |'''
     if len(p) == 1:
-        p[0] = threadlocal.field_seq_implicit_id.get_id()
+        p[0] = next(threadlocal.field_seq_implicit_id)
     else:
         p[0] = p[1]
 
@@ -570,7 +571,7 @@ def parse(path, module_name=None, include_dirs=None, include_dir=None,
         threadlocal.include_dirs_ = ['.']
         threadlocal.thrift_cache = {}
         threadlocal.incomplete_type = CurrentIncompleteType()
-        threadlocal.field_seq_implicit_id = CurrentFieldSeqImplicitId()
+        threadlocal.field_seq_implicit_id = itertools.count(start=-1, step=-1)
         threadlocal.initialized = True
 
     # dead include checking on current stack
@@ -657,7 +658,7 @@ def parse_fp(source, module_name, lexer=None, parser=None, enable_cache=True):
         threadlocal.include_dirs_ = ['.']
         threadlocal.thrift_cache = {}
         threadlocal.incomplete_type = CurrentIncompleteType()
-        threadlocal.field_seq_implicit_id = CurrentFieldSeqImplicitId()
+        threadlocal.field_seq_implicit_id = itertools.count(start=-1, step=-1)
         threadlocal.initialized = True
 
     if not module_name.endswith('_thrift'):
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -375,8 +375,13 @@ def p_field(p):
 
 
 def p_field_id(p):
-    '''field_id : INTCONSTANT ':' '''
-    p[0] = p[1]
+    '''field_id : INTCONSTANT ':'
+                |'''
+    if len(p) == 1:
+        # TODO: How do we keep state to decrement this for every field missing and id?
+        p[0] = -1
+    else:
+        p[0] = p[1]
 
 
 def p_field_req(p):
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -348,6 +348,7 @@ def p_field_seq(p):
     '''field_seq : field sep field_seq
                  | field field_seq
                  |'''
+    threadlocal.incomplete_struct_fields = CurrentIncompleteStructFields()
     _parse_seq(p)
 
 
@@ -378,7 +379,7 @@ def p_field_id(p):
     '''field_id : INTCONSTANT ':'
                 |'''
     if len(p) == 1:
-        p[0] = threadlocal.incomplete_type.set_info((None, p.lineno(0)))
+        p[0] = threadlocal.incomplete_struct_fields.set_info((None, p.lineno(0)))
     else:
         p[0] = p[1]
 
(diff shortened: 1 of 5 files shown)


aisk:
Hi, thank you for your contribution. 

To my surprise, this is the first time I've heard that the field ID is optional (I'm not the original author of the Thrift parser in ThriftPy).

As far as I know, ThriftPy2 has been used by some large companies with tons of Thrift IDL files. There are also many open-source projects on GitHub: https://github.com/search?q=+language%3Athrift&type=code&p=1, and none of them seem to use Thrift with field IDs omitted.

The official Thrift documentation you referenced above doesn't even mention that field IDs can be omitted—except in the BNF grammar. The same goes for other resources like [this guide](https://diwakergupta.github.io/thrift-missing-guide/) and [this one](https://thrift-tutorial.readthedocs.io/en/latest/thrift-types.html).

So I think this might be a deprecated feature that is not widely used, and it might be better **not** to implement it in Thriftpy.

Just my personal opinion—I'd like to wait for input from other maintainers. I'm also open to hearing other reasons in favor of supporting this.

Fluxx:
Thanks for the comment. For some context on my end, at my job we have a few services who have omitted their ids for some structs and service definition, and those are in use in production, and in some cases have been for several years. I do not think omitting the ids was intentional on their part, likely someone just did it as an accident. We use [Scrooge](https://github.com/twitter/scrooge) as our Thrift compiler for Scala, and also the official `thrift` compiler for Ruby, and both of those compilers worked with this accident without any issue. We only realized we had this problem because I am doing work now to integrate Python with our Thrift ecosystem, and ran into an error.

I do agree that omitting the id is not a good idea, is not widely known or used - even within the official Thrift docs, and would not recommend anyone actually omit their ids. In an ideal world we'd have our services who omitted their id add an id back, but doing so is somewhat tedious for reasons I mentioned in the PR description. Tedium is not a reason in and of itself to not do it, but given the Thrift spec seems to indicate it's allowed, and the official `thrift` compiler also allows it, I think it's a reasonable argument that is should be supported.

Curious to hear what the other maintainers of this repo say as well.

ethe:
Considering the patch size, I'm open to accept the PR if all cases in pytest are passed.

Fluxx:
I believe all the tests are passing. The failing checks all appear to be lint failures in code I didn't update; but I can fix that lint error as part of this PR if you'd like?

thriftpy2/parser/parser.py
```diff
@@ -404,6 +409,16 @@ def set_info(self, info):
         return self.index + 1
 
 
+class CurrentFieldSeqImplicitId(dict):
```
aisk, line 412: I think this is an anti-pattern in OOP. It's not a dict and doesn't behave like one, so it shouldn't be a subclass of dict. You can just let it inherit from object directly and use a dict as an instance variable.

thriftpy2/parser/parser.py
```diff
@@ -404,6 +409,16 @@ def set_info(self, info):
...
+    def __init__(self, *args, **kwargs):
+        self.index = -1
+
+    def set_info(self, info):
+        self[self.index] = info
```
aisk, line 417: Why store the *info* in this object? Seems it doesn't accessed by any other place.

thriftpy2/parser/parser.py
Fluxx: Sure thing, can change. For context this was modeled off the `CurrentIncompleteType` class that already existed in this module, and does inherit from `dict`:

https://github.com/Thriftpy/thriftpy2/blob/56c6df707e282d6f1bf795667dda27c12af36b56/thriftpy2/parser/parser.py#L403-L409

thriftpy2/parser/parser.py
Fluxx: This was also modeled off the `CurrentIncompleteType`, which stored the `info`. It wasn't 100% clear to me why that class also stored the info, but given I am unfamiliar with this library, ply, and lex and yacc in general, I just followed the existing pattern.

 I can see if I can remove the storage of the info here, and just use the class to track the index.

Fluxx:
@aisk alright, I simplified the `CurrentFieldSeqImplicitId` class to not inherit from dict and also remove setting the info. Please take a look.

ethe:
`CurrentIncompleteType` stores a map about types that have been parsed but not yet found, to their positions in the ast tree, I think the case here is different.

Fluxx:
> `CurrentIncompleteType` stores a map about types that have been parsed but the difinition not yet be found, to their positions in the ast tree, I think the case here is different.

Ah, that's good to know. Thanks for the context.

cocolato:
It might be a better choice to replace `CurrentIncompleteType`  with  [`itertools.count(start=0, step=-1)` ](https://docs.python.org/3.13/library/itertools.html#itertools.count).

Fluxx:
> It might be a better choice to replace `CurrentIncompleteType` with [`itertools.count(start=-1, step=-1)` ](https://docs.python.org/3.13/library/itertools.html#itertools.count).
> 
> ```
> >>> import itertools
> >>> counter = itertools.count(start=-1, step=-1) 
> >>> next(counter)
> -1
> >>> next(counter)
> -2
> ```

I liked this idea, and was one I had thought of yesterday just to simplify the tooling to decrement the id. I just made that change.

aisk:
Except the change for `itertools.count`, this PR looks good to me.

Fluxx (commented):
Oh sorry everyone, I thought I pushed my changed last week, but I did not. I just pushed them.

aisk (approved):

thriftpy2/parser/parser.py
```diff
@@ -404,6 +410,15 @@ def set_info(self, info):
...
 
 
+class CurrentFieldSeqImplicitId(object):
+    def __init__(self):
+        self.index = -1
+
+    def get_id(self):
+        self.index -= 1
+        return self.index + 1
+
```
cocolato, line 420: `CurrentFieldSeqImplicitId` should be able to be completely removed.

Fluxx: Good idea. I just pushed a commit to do that.

cocolato (approved):

Fluxx:
Thanks everyone! Glad to get this fix merged in.

Issue #294, opened 2024-09-19
How could we get namespace values from thriftpy2.parser


truebit:
I saw two methods in parser.py concerning `namespace`:

```
def p_namespace(p):
    '''namespace : NAMESPACE namespace_scope IDENTIFIER'''
    # namespace is useless in thriftpy2
    # if p[2] == 'py' or p[2] == '*':
    #     setattr(thrift_stack[-1], '__name__', p[3])


def p_namespace_scope(p):
    '''namespace_scope : '*'
                       | IDENTIFIER'''
    p[0] = p[1]
```
But I cannot find namespace attributes after parsing the IDL file using `from thriftpy2.parser.parser import parse`

PR #308, opened 2025-04-22
add namespace attributes to __thrift_meta__, then we could access namespace info. 

truebit:
add namespace attributes to `__thrift_meta__`, so we could access namespace info. support #294 

```
thrift = load("some.thrift")
namespaces = thrift.__thrift_meta__['namespaces']
cpp_namespace = namespaces.get('cpp')  # get ns by cpp lang
java_namespace = namespaces.get('java')  # get ns by java lang
```

### DIFF
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -90,9 +90,23 @@ def p_cpp_include(p):
 
 def p_namespace(p):
     '''namespace : NAMESPACE namespace_scope IDENTIFIER'''
-    # namespace is useless in thriftpy2
-    # if p[2] == 'py' or p[2] == '*':
-    #     setattr(threadlocal.thrift_stack[-1], '__name__', p[3])
+    thrift = threadlocal.thrift_stack[-1]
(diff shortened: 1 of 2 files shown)


thriftpy2/parser/parser.py
```diff
@@ -90,9 +90,23 @@ def p_cpp_include(p):
...
+        setattr(thrift, '__thrift_meta__', meta)
+    else:
+        meta = getattr(thrift, '__thrift_meta__')
+    
+    # Initialize 'namespaces' as dict if not exists
+    if 'namespaces' not in meta:
+        meta['namespaces'] = {}
+    
+    # Store namespace information
+    meta['namespaces'][scope] = namespace
```
cocolato, line 109:
```suggestion
     _add_thrift_meta("namespaces", (p[2], p[3]))
```
I don't think namespaces should be a one - way map. Implementing it in this way would be more reasonable.

cocolato (changes requested):

aisk:
Thank you for the contribution. This has now been addressed by #370, so I’m closing this PR in favor of that implementation.

Issue #318, opened 2025-07-15
0.5.3 added dependency on unmaintained `toml` package


mgorny:
0.5.3 started using the [toml](https://pypi.org/project/toml/) package that has been abandoned in 2020 and doesn't implement TOML v1.0.0. Please use [tomllib](https://docs.python.org/3/library/tomllib.html) from stdlib (3.11+), with fallback to [tomli](https://pypi.org/project/tomli/).

PR #319, opened 2025-07-15
remove abandoned toml dependency

cocolato:
Fix: https://github.com/Thriftpy/thriftpy2/issues/318
Use [tomli](https://pypi.org/project/tomli/) and [tomllib](https://docs.python.org/3/library/tomllib.html) to replace abandoned [toml](https://pypi.org/project/toml/) package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=68",
     "cython>=0.28.4,<4",
-    "tomli; python_version < '3.11'"
+    "tomli; python_version<'3.11'"
 ]
 
 [project]


cocolato:
> Maybe we can just use the tomllib in standard library with Python >= 3.11?

It has been updated.

pyproject.toml
```diff
@@ -1,5 +1,9 @@
 [build-system]
-requires = ["setuptools>=68", "cython>=0.28.4,<4", "tomli"]
+requires = [
+    "setuptools>=68",
+    "cython>=0.28.4,<4",
+    "tomli; python_version < '3.11'"
```
aisk, line 5:
```suggestion
    "tomli; python_version<'3.11'"
```

Issue #323, opened 2025-09-18
Local struct shadows imported module


FrankPortman:
When a local struct has the same name as an imported module (e.g., local struct
named 'user_types' when there's an imported module 'user_types.thrift'), thriftpy2
incorrectly resolves ALL qualified references like 'user_types.UserProfile' to
the local struct instead of accessing the struct from the imported module.

This makes it impossible to reference structs from an imported module when there's
a naming collision with a local struct.

EXPECTED BEHAVIOR:
- 'user_types.UserProfile' should resolve to UserProfile from user_types module
- 'user_types' (unqualified) should resolve to the local struct

ACTUAL BEHAVIOR:
- Both resolve to the local 'user_types' struct
- The qualified path is completely ignored

Ref files/tests:

```thrift
// user_types.thrift
struct UserProfile {
    1: required string user_id
    2: required string username
    3: optional string email
}
```

```thrift
// main.thrift

include "user_types.thrift"

struct UserProfile { 
    1: required string local_field1
    2: optional i32 local_field2
}

struct ApplicationData {
    // Case 1a: Reference to local struct that has same name as imported module
    1: required user_types localUserTypes

    // Case 1b: SHOULD reference user_types.UserProfile from imported module
    // BUT will incorrectly resolve to local user_types struct
    2: required user_types.UserProfile importedUserProfile
}

```

However it is okay (and it is expected to be okay) if the imported `struct` collides with the name of a local `struct`. It seems the issue only occurs if the imported `module` collides with the name of a local `struct`.

aisk:
Hi @FrankPortman, thank you for reporting this issue, and sorry for the late reply.

I want to use the code in your PR's test as an example. Let's say we have two thrift files:

`name_collision_imported.thrift`:

```thrift
// This file is imported and has a struct named UserProfile
struct UserProfile {
    1: required string user_id
    2: required string username
    3: optional string email
}
```

`name_collision.thrift`:

```thrift
// Test case for when a local struct has the same name as an imported module
include "name_collision_imported.thrift"

// Local struct with the same name as the imported module
struct name_collision_imported {
    1: required string local_field1
    2: optional i32 local_field2
}

struct TestStruct {
    // Should resolve to the local struct 'name_collision_imported'
    1: required name_collision_imported localStruct

    // Should resolve to UserProfile from the imported module
    2: required name_collision_imported.UserProfile importedUserProfile
}
```

[...]

PR #324, opened 2025-09-18
fix: local struct shadows included module

FrankPortman:
fixes: https://github.com/Thriftpy/thriftpy2/issues/323


FrankPortman:
@aisk not sure if you are the current maintainer but I see you have the most commits in the last half year so please take a look when you can 🙏

FrankPortman:
Friendly bump - @cocolato tagging you as well since I see some recent activity from you. 

Some initial direction to this PR or the underlying issue would be great!

cocolato:
https://github.com/Thriftpy/thriftpy2/blob/8e82f1c091a9d9db30dc01af25431d06056c3d88/thriftpy2/parser/parser.py#L82-L84 https://github.com/Thriftpy/thriftpy2/blob/8e82f1c091a9d9db30dc01af25431d06056c3d88/thriftpy2/parser/parser.py#L244-L248 The root cause appears to be that the p_seen_struct method overwrites the contents of the originally imported module via setattr. Although the current fix resolves this issue, I suggest using a prefix with setattr to better distinguish between the imported module and the struct. @aisk  please take a look.

Issue #278, opened 2024-06-20
how to get the annotation of service function?


helloyuqin:
could I get the annotation?

helloyuqin:
@aisk could you help me to solve this issue？

helloyuqin:
突然发现原来是安龙大佬，求助大佬

aisk:
Currently, thriftpy doesn't provide this feature. Can you please describe your use case for the type annotations? If it's useful, maybe we can add it now.

kiwiz:
+1 on this. It'd be useful for me to determine if a Thrift struct/service/etc has an annotation I expect.

PR #336, opened 2026-01-07
Add annotation support

jlamberts-yobi:
Adds support for most annotations. For our use case, we actually really just need field-level annotations, but I took a look at the issues and saw others were interested in broader annotations in #278, so figured I'd include those as well.

Specifically:
- Struct-level annotations get a `__thrift_annotations__` attr, with struct field annotations under the `__thrift_field_annotations__` attr
- Enum-level annotations get a `__thrift_annotations__` attr, with item annotations under `__thrift_item_annotations__`
- Service-level annotations get a `__thrift_annotations__` attr, with function annotations under `__thrift_function_annotations__`
- The top-level module gets **lazy**  typedef **alias** and const annotations under `__thrift_typedef_annotations__` and `__thrift_const_annotations__`, respectively

Note that this PR **does not** add support for annotations on the types, mostly because that would touch a lot more code: since types are represented by ints, we can't easily add metadata (like annotations). 

So we parse the "foo=bar" annotation, **but not the unicode.encoding** annotation in `typedef string ( unicode.encoding = "UTF-16" ) non_latin_string (foo="bar")`.

### DIFF
--- a/thriftpy2/parser/parser.py
+++ b/thriftpy2/parser/parser.py
@@ -393,7 +393,7 @@ def p_simple_field(p):
 
 def p_field(p):
     '''field : simple_field type_annotations'''
-    p[0] = p[1] + [p[2]] 
+    p[0] = p[1] + [p[2]]
 
 
 def p_field_id(p):


aisk:
Hi @jlamberts-yobi, thank you for your contribution. This feature and the code quality look very good at first glance.

However, I’m quite busy these days, so I can’t finish the review right now. I plan to complete it this weekend or early next week, so please be patient.

And let’s also see the review comments from other people during this period.

cocolato:
Thanks! LGTM.

PR #352, opened 2026-03-09
Improve Apache JSON framing performance by using `JSONDecoder.raw_decode`

aisk:


aisk:
This PR tried to decode the JSON from the beginning for every chunk received, not incrementally. I found that we can parse the JSON stream incrementally using ijson, so I will close this.

PR #354, opened 2026-03-25
Add Windows ARM64 support in CI

Harishmcw:
This PR adds Windows ARM64 support to the CI test and build matrix.

- Adds `windows-11-arm` to the os matrix in both Test and Build workflows.
- In Test, excludes Python 3.8, 3.9, 3.10 and all PyPy versions as they are not supported on Windows ARM64.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


cocolato (approved):

ethe (approved):

aisk:
There is no need to release a new version because there is no actual changes from last release.

I just uploaded the Arm64 on Windows wheels to the exist `v0.6.0`, you can install them now: https://pypi.org/project/thriftpy2/0.6.0/#files

PR #356, opened 2026-04-09
Write buffer protocol

vallsv:
Hi,

This pull request features the `write` binary support for the Python buffer protocol.

As result, on top of the already supported `str` and `bytes`, the API now supports `bytearray` and `memoryview`. Everything use the same new Cython `write_buffer` function. Pure python writer was also implemented.

This can be used to reduce a memory copy while using numpy arrays.

I don't have checked the result with a complex benchmark, but in my use case (arrays of about 16MB), i can notice a time reduction of about 10% in the average (client and server on the same machine). For small messages, this does not change anything. I don't really know what could be a good benchmark.

Actually i use it such way:

```
struct NDArray {
    1: string dtype;
    2: list<i32> shape;
    3: binary buffer;
}
```

```
def create_ndarray_record(array: numpy.ndarray):
    msg = NDArray()
    msg.shape = array.shape
    msg.dtype = array.dtype.str

    # msg.buffer = array.tobytes()  # <-- previous code, this creates a memory copy

    array = numpy.ascontiguousarray(array)
    msg.buffer = array.data

    return msg
```

The pull request does not have any dependency to `numpy`.

Would you be interested by such contribution? Tell me the way i could improve it.

### DIFF
--- a/thriftpy2/protocol/binary.py
+++ b/thriftpy2/protocol/binary.py
@@ -121,8 +121,9 @@ def write_val(outbuf, ttype, val, spec=None):
     elif ttype in BIN_TYPES:
         if isinstance(val, str):
             val = val.encode('utf-8')
-        val = bytes(val)
-        outbuf.write(pack_string(val))
+        val = memoryview(val)
+        outbuf.write(pack_i32(val.nbytes))
+        outbuf.write(val)
 
     elif ttype == TType.SET or ttype == TType.LIST:
         if isinstance(spec, tuple):
--- a/thriftpy2/protocol/binary.py
+++ b/thriftpy2/protocol/binary.py
@@ -119,8 +119,9 @@ def write_val(outbuf, ttype, val, spec=None):
         outbuf.write(pack_double(val))
 
     elif ttype in BIN_TYPES:
-        if not isinstance(val, bytes):
+        if isinstance(val, str):
             val = val.encode('utf-8')
+        val = bytes(val)
         outbuf.write(pack_string(val))
 
     elif ttype == TType.SET or ttype == TType.LIST:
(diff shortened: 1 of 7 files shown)


aisk:
The idea sounds reasonable to me. However, we should just support [buffer protocol](https://docs.python.org/3/c-api/buffer.html) instead of just support memoryview. As a result, `memoryview` / `numpy.array` / `torch.tensor` both implemented buffer protocol, they can be used in thriftpy2 directly.

vallsv:
@aisk thanks for your feedback. Here is a rework with the buffer protocol.

I have also added few tests to check if both binary/cybinary can serialize few objects exposing the buffer protocol.

thriftpy2/protocol/binary.py
```diff
@@ -119,8 +119,9 @@ def write_val(outbuf, ttype, val, spec=None):
         outbuf.write(pack_double(val))
 
     elif ttype in BIN_TYPES:
-        if not isinstance(val, bytes):
+        if isinstance(val, str):
             val = val.encode('utf-8')
+        val = bytes(val)
```
aisk: This can avoid memory copy if val object support, and make the behavior more like the cython version.

```suggestion
        val = memoryview(val)
```

vallsv:
`struct.pack` is not supporting `memorymap`.

It is possible to pack it like that.
```
def pack_memoryview(mem):
    return struct.pack("!i", len(mem)) + mem
```

But I have choose to write it in 2 steps instead. I guess it's better.

vallsv:
@aisk here the problem is that the pure python `binary` module is using `TCyMemoryBuffer`.

That's in between 2 worlds.

I am thinking if it makes sens to propagate that at such place? Ill take a look anyway.

aisk:
> @aisk here the problem is that the pure python `binary` module is using `TCyMemoryBuffer`.
> 
> That's in between 2 worlds.
> 
> I am thinking if it makes sens to propagate that at such place? Ill take a look anyway.

 I think it's OK to figure this out in the future.

aisk (approved):

PR #379, opened 2026-07-25
fix: allow enum default value declared after its use

### LINKED ISSUE #205: Default value with enum must be declared before used
``` thrift
struct A {
    1: Type t = Type.AA
}

enum Type {
    AA = 1
    BB = 2
}
```
And we got error **thriftpy2.parser.exc.ThriftParserError: Cann't find name 'Type.AA' at line 2**

``` thrift
enum Type {
    AA = 1
    BB = 2
}

struct A {
    1: Type t = Type.AA
}
```
This will be fine.

Sanjays2402:
Closes #205

A field default that references an enum member defined *later* in the same thrift file raised `ThriftParserError: Can't find name 'Type.AA'`, even though forward references already work for field types (via `incomplete_type` + the `fill_incomplete_ttype` second pass).

`p_const_ref` now defers an unresolved name as an `UNKNOWN_CONST_REF` marker instead of raising immediately, and `fill_incomplete_ttype` resolves it — regenerating the struct's `default_spec`/`__init__` — once the whole file has been parsed. A genuinely undefined name still raises, just from the second pass.

New `test_enum_default_value_declared_after_use` fails with the reported error without the fix and passes with it; `tests/test_parser.py` is otherwise unchanged (41 passed, same 2 pre-existing thread-test errors before and after).

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -478,3 +478,24 @@ def test_thrift_in_include_path():
 
 def test_issue_121():
     load(TEST_DIR / 'parser-cases/issue_121.thrift')
+
+
+def test_enum_default_value_declared_after_use(tmp_path):
+    """Regression: an enum used as a field default before it is declared.
+
+    Forward references are already supported for field *types*; a default
+    value referring to a not-yet-parsed enum member used to raise
+    ThriftParserError("Can't find name 'Type.AA'").
+    """
+    path = tmp_path / 'forward_enum_default.thrift'
+    path.write_text(
+        'struct A {\n'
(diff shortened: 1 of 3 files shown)


aisk:
This only works when the field type is also a forward reference, because `p_simple_field` and `p_const` still cast eagerly. In every other position the placeholder tuple either fails with a misleading "Type error" or leaks into the generated defaults, and real typos are no longer rejected.

Better to defer the cast until after `parser.parse()` returns instead of adding a second sentinel next to `UNKNOWN_CONST`.

PR #380, opened 2026-07-25
fix: resolve include statements in load_fp when source is a real file

Sanjays2402:
Closes #285

`parse_fp` always set `__thrift_file__` to `None`, so `p_include` had no base directory to resolve against and rejected any thrift file with an `include` statement — even when the file-like object was an ordinary `open()` file whose path is known. When `source` has a usable `.name` pointing at an existing file, it is now used as `__thrift_file__` so includes resolve relative to it, exactly as `parse()` does; non-file objects (`StringIO`, sockets) keep the existing error.

`test_load_fp_with_include` fails on master with the reported `ThriftParserError` and passes with the fix; `test_e_load_fp` now covers the still-unresolvable case via `StringIO`.

PR #381, opened 2026-07-25
fix(parser): allow include statements in load_fp

### LINKED ISSUE #285: `thriftpy2.load_fp` can't load a thrift file with a include statement
In the project's tests directory:

```
>>> import thriftpy2
>>> f = open('addressbook.thrift')
>>> thriftpy2.load_fp(f, 'addressbook_thrift')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/asaka/Codes/thriftpy2/thriftpy2/parser/__init__.py", line 171, in load_fp
    thrift = parse_fp(source, module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/asaka/Codes/thriftpy2/thriftpy2/parser/parser.py", line 655, in parse_fp
    parser.parse(data)
  File "/Users/asaka/Codes/thriftpy2/venv/lib/python3.12/site-packages/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/asaka/Codes/thriftpy2/venv/lib/python3.12/site-packages/ply/yacc.py", line 1120, in parseopt_notrack
    p.callable(pslice)
  File "/Users/asaka/Codes/thriftpy2/thriftpy2/parser/parser.py", line 58, in p_include
    raise ThriftParserError('Unexpected include statement while loading '
thriftpy2.parser.exc.ThriftParserError: Unexpected include statement while loading from file like object.
```

Sanjays2402:
Closes #285

`load_fp`/`parse_fp` rejected any thrift source containing an `include`, so opening a file and passing it to `load_fp` raised `Unexpected include statement while loading from file like object.` even when the included files were resolvable.

A module parsed from a file-like object has no path of its own, so `p_include` can't derive a base directory from `__thrift_file__`; it now falls back to the parse context's `include_dirs` (default `['.']`) and names the child module from the included file's basename. The file-backed path is unchanged.

`tests/test_parser.py::test_load_fp_with_include` fails with the reported error without the parser change and passes with it.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -338,6 +338,24 @@ def test_load_fp():
     assert thrift.__thrift_meta__['services'] == [thrift.SharedService]
 
 
+def test_load_fp_with_include(monkeypatch):
+    from thriftpy2.parser import parser as _parser
+    _parser._thrift_cache.clear()
+
+    monkeypatch.chdir(TEST_DIR / 'parser-cases')
+    try:
+        with open('include.thrift') as thrift_fp:
+            thrift = load_fp(thrift_fp, 'include_fp_thrift')
+
+        assert thrift.__thrift_file__ is None
+        assert thrift.datetime == 1422009523
+        assert [t.__name__ for t in thrift.__thrift_meta__['includes']] == \
+            ['included', 'included_1']
+    finally:
+        # the includes were cached under paths relative to this directory
+        _parser._thrift_cache.clear()
+
+
 def test_e_load_fp():
     with pytest.raises(ThriftParserError) as excinfo:
         with open(TEST_DIR / 'parser-cases/tutorial.thrift') as thrift_fp:
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import threading
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


aisk:
Child module names for fp-loaded roots are the bare basename, and `_thrift_cache` is keyed by module name. Including `a/t.thrift` and `b/t.thrift` from two different roots both map to `t_thrift`, so the second one silently gets the first file's module. The same file also gets cached under different keys depending on whether the parent came from `load` or `load_fp`, so you end up with two distinct struct classes.

`load_fp` also doesn't register included children in `sys.modules` like `load()` does. Structs from an included file point at a module that can't be imported, so pickling them fails.

Adding `include_dirs` to `parse_fp` / `load_fp` and computing the child name with `os.path.relpath` in both branches would fix both, and drop the cwd dependency that currently forces `monkeypatch.chdir` in the test.

Also please fix the typecheck failure in CI. pyright complains about `os.path.basename` at `parser.py:81` getting a possibly `None` argument.

Sanjays2402:
All addressed in 69ff35e:

- **Child name collisions**: the fp branch now computes the child module name with `os.path.relpath(path, include_dir)` (anchored at the include dir the child was resolved from) instead of the bare basename, so including `a/t.thrift` and `b/t.thrift` from two fp-loaded roots yields `a.t_thrift` / `b.t_thrift`. Added `test_load_fp_include_same_basename_no_collision`.
- **`load` vs `load_fp` divergence**: since the same child file now gets the same relpath-derived module name in both branches, it hits the same `_thrift_cache` key and produces the same classes. Added `test_load_fp_include_child_matches_load` asserting `fp_thrift.t.A is file_thrift.t.A`.
- **`sys.modules` registration**: `load_fp` now registers included children recursively (extracted the loop from `load()` into `_register_include_modules`, used by both, including the name-conflict check). Added `test_load_fp_include_child_picklable` — a struct from an included file round-trips through pickle.
- **`include_dirs`**: added to both `parse_fp` and `load_fp`, plumbed into the parse context like `parse()` does. The test no longer needs `monkeypatch.chdir` — it passes `include_dirs=[TEST_DIR / 'parser-cases']`.
- **pyright**: `thrift.__thrift_file__` is now bound to a local before the `None` check so the narrowing applies to the `os.path.basename` call.

Tests: `tests/test_parser.py` is 45 passed (the 2 errors are the pre-existing `test_load_in_sub_thread`/`test_load_fp_in_sub_thread` missing-`reraise`-fixture errors, also present without this change). Note #380 is still open — I'll leave the pick-one-and-close-the-other call to Sanjay.

aisk:
> Note https://github.com/Thriftpy/thriftpy2/pull/380 is still open — I'll leave the pick-one-and-close-the-other call to Sanjay.

This reads like an AI replying to my review on your behalf. Using AI to write code is fine in this project, but replying to reviewer comments with AI is not, other than for proofreading or translation. If that's what happened here, I'll close these PRs.

